PR #2, opened 2021-06-07
Add a deriving option to generate NSCopying implementation for ObjC records

finalpatch:
For client code that expects NSCopying conforming objects, this adds support for using deriving (nscopying) on records to generate a minimal implementation (simple retain) of NSCopying in ObjC. Since all djinni records are immutable, retaining the object is enough. Java and C++ code generation are not affected.


benjaminhollissc (approved):
Neat! One question I'd have is if there's a more cross-platform concept that could be used for this than `nscopying`, which could be then implemented across other outputs as well. e.g. `Eq` is not specific to any one platform.

finalpatch:
> Neat! One question I'd have is if there's a more cross-platform concept that could be used for this than `nscopying`, which could be then implemented across other outputs as well. e.g. `Eq` is not specific to any one platform.

I thought about naming it just "copy" or "clone" but Java's Clonable has different semantic than NSCopying. Simple retaining is a legit way to implement NSCopying for immutable types (Apple document explicitly says this), but Java Clonable is intended for making a deep copy so that the 2 copies can be modified independently. And then C++ just has its default copy constructor that's always there. I decided to use just "nscopying" to make it clear that this is an ObjC only thing because we don't need this in the other languages. If we don't need them, we don't have to invent that need just for the sake of the name.

If we later find this concept can be extended to other languages in a meaningful way, we can easily change the name and make "nscopying" an alias to the new name.

PR #7, opened 2021-09-02
Generate string names for C++ enums

finalpatch:
This generates a `to_string()` function for non-flag enums in C++.   flag enums are unchanged for now because creating strings for them has higher overhead and involves allocation.

Testing: one of the TestHelpers functions from the unit test suite is modified to exercise the new to_string() function.


examples/generated-src/cpp/sort_order.hpp
```diff
@@ -13,6 +13,8 @@ enum class sort_order : int {
     RANDOM,
 };
 
+const char* to_string(sort_order e) noexcept;
```
bholmes-sc, line 16: can we make these constexpr?

finalpatch: changed

bholmes-sc (approved):

PR #8, opened 2021-09-16
Add Web Assembly support

finalpatch:
This PR adds fairly complete support for web assembly:

* Wasm code generator
* Wasm support library
* TypeScript generator
* Test suites in plain Javascript and TypeScript
* Example web app
* PerfTest web app

Bazel build integration is not yet implemented so Wasm code is currently built with Makefiles.

Almost all Djinni features are supported, including custom type translators.

Notable differences when comparing to the Java/ObjC support:
* deriving(ord, eq, hash) is not applicable to Javascript because JS doesn't support overloading standard comparison methods.
* Extended records generate the same code as regular records. Because JS can easily add extension methods (by add functions to prototype) without having to derive from a base class.
* Protobuf support is incomplete (mainly because we don't have a wasm build of protobuf runtime, will add this after bazel build integration)


benjaminhollissc (commented):
Cool! Let's be clear somewhere that this is experimental/unstable and still subject to changes.

bkotsopoulossc (approved):

support-lib/cpp/DataRef.cpp
```diff
@@ -324,6 +329,89 @@ DataRef::DataRef(CFDataRef platformObj) {
     _impl = std::make_shared<DataRefObjc>(platformObj);
 }
 
+#elif DATAREF_WASM
+
+class DataRefWasm : public DataRef::Impl {
```
bholmes-sc, line 334: can we just put this into a separate file and include it when needed?   very large #if blocks seems problematic to maintain over time.

bholmes-sc (approved):

PR #9, opened 2021-10-01
[Bazel] Allow consumption of repo as bazel dependency

bkotsopoulossc:
Even though the project currently is building with Bazel, there currently is no support to consume it as a dependency via bazel. The convention to doing this in bazel is to move all dependency setup into a function that can be called from consuming projects. However, many dependencies require doing a `load(..., some_func)` and then calling the function `some_func()` to set things up, and bazel has a restriction that you can only `load()` at the top of a .bzl extension, or in `WORKSPACE`. The typical way to work around this is to do a two-step setup, where the first step fetches all dependencies, and the second step loads and calls relevant functions to set them up. We can change nomenclature if anyone feels particularly strongly.

Once this is merged, we can add the following setup instructions to the readme:
```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

djinni_commit = "..."
http_archive(
    name = "djinni",
    sha256 = "...",
    strip_prefix = "djinni-{}".format(djinni_commit),
    type = "zip",
    url = "https://github.com/Snapchat/djinni/archive/{}.zip".format(djinni_commit),
)

load("@djinni//bzl:deps.bzl", "djinni_deps")
djinni_deps()

load("@djinni//bzl:setup_deps.bzl", "djinni_setup_deps")
djinni_setup_deps()
```

At that point, the compiler should be runnable with: `bazel run @djinni//bzl:djinni --idl ...`


bkotsopoulossc:
PTAL @LiFengSC @bholmes-sc CC @benjaminhollissc

bzl/deps.bzl
```diff
@@ -0,0 +1,78 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+def djinni_deps():
+
+    native.android_sdk_repository(name = "androidsdk")
```
bholmes-sc, line 5: i would imagine this particular setting/dep would be separate from other project deps.

bholmes-sc (approved):

bzl/deps.bzl
```diff
@@ -0,0 +1,78 @@
...
+    native.android_ndk_repository(name = "androidndk", api_level = 21)
+
+    # Bazel Skylib needed for Scala rules
+    bazel_skylib_version = "1.0.2"
+    http_archive(
```
chancila, line 10: to be good citizens here we should wrap these call in maybe in case the root project already set these deps up...

https://docs.bazel.build/versions/main/repo/utils.html#maybe

bzl/deps.bzl
bkotsopoulossc: Yeah I wasn't quite sure exactly how to model this. Maybe this lives in WORKSPACE and we don't consider it a dependency

bkotsopoulossc:
Also my builds aren't running without extra permissions
image

bzl/setup_deps.bzl
```diff
@@ -0,0 +1,33 @@
...
+load("@io_bazel_rules_kotlin//kotlin:kotlin.bzl", "kotlin_repositories", "kt_register_toolchains")
+
+def djinni_setup_deps():
+
+    bazel_skylib_workspace()
+
+    scala_register_toolchains()
+    scala_repositories()
+
+    maven_install(
```
chancila, line 17: this really should probably have an explicit name in case a top level project already uses the default `mvn`...

probably not for this patch though @LiFengSC

bzl/deps.bzl
chancila, line 5: yea these android and swift/apple/kotlin examples probably should probably have their own WORKSPACES instead of being part of the main workspace, otherwise you're polluting with extra deps for no reason @LiFengSC

finalpatch: Yes.  I didn't know we can have nested WORKSPACEs

chancila: you cannot, but typically example workspaces are `.bazelignore`'ed and set up to be standalone and inner workspaces are made to work with `local_repository` and relative paths to the top level one:

https://github.com/bazelbuild/rules_kotlin/blob/master/.bazelignore
https://github.com/bazelbuild/rules_kotlin/blob/master/examples/android/WORKSPACE#L5

bkotsopoulossc: I'll move these back to the top-level WORKSPACE for now

finalpatch (approved):
Looks good

.github/workflows/build.yaml
```diff
@@ -30,3 +30,7 @@ jobs:
 
         - name: Test
           run: bazel test //test-suite:all
+
+        - name: External Test
+          working-directory: external-test
+          run: bazel run @djinni//src:djinni -- --help
```
bkotsopoulossc, line 36: Adding a simple test here to ensure we can consume the compiler from an external workspace. This seems trivial but it will fail if we can't build and/or run

WORKSPACE
```diff
@@ -1,107 +1,59 @@
...
+# --- Everything below is only used for examples and tests
+
 android_sdk_repository(name = "androidsdk")
 android_ndk_repository(name = "androidndk", api_level = 21)
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
-load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
-# Bazel Skylib needed for Scala rules
+build_bazel_rules_apple_version = "0.31.3"
```
bkotsopoulossc, line 16: I've moved all of the android and iOS deps back to WORKSPACE since we don't need it in the dependencies exposed externally

external-test/README.md
```diff
@@ -0,0 +1,6 @@
+External Test
+-------------
+This folder provides a standalone workspace to mimic consuming Djinni from an external repository via Bazel.
+For now, this simply tests that we can build and run the compiler. Eventually a more complete setup will be
+exercised here. To run, navigate to this folder and then use:
```
bkotsopoulossc, line 5: A more complete setup for this external test would be using rules for codegen, with linking against the support lib.

PR #20, opened 2021-10-19
Proposal: Build-time check for use of JS interfaces without an impl

bkotsopoulossc:
If you have a djinni file, and run the codegen, but forgot to set the appropriate language flags (`+w, +c, +j, +o`) on a specific interface, the translation layer falls back to an empty value, and if you attempt to map that type through the language boundary, it blows up at runtime. This feels like we have an opportunity to provide safeguards here, and we aren't.

Proposal is to have a build time check to make sure the fallback methods are never used. We can use SFINAE if anyone feels strongly, but I prefer explicit error messages.

Note that I'm not sure if we rely on this fallback behaviour anywhere. If we do, I feel like there are strong reasons not to.

Thoughts @LiFengSC

### DIFF
--- a/support-lib/wasm/djinni_wasm.hpp
+++ b/support-lib/wasm/djinni_wasm.hpp
@@ -458,7 +458,13 @@ struct JsInterface {
     // (interface +c)
     template <typename, typename>
     struct GetOrCreateCppProxy {
-        em::val operator() (const std::shared_ptr<I>& c) { return em::val::undefined(); }
+        em::val operator() (const std::shared_ptr<I>& c) {
+            // Trying to create a C++ object from JS without a C++ impl will just give undefined.
+            // This is probably not what you want. Either expose the interface to C++ and implement it,
+            // or make sure this method does not end up being called.
+            static_assert(false, "attempting to create c++ proxy object for interface without +c");
+            return em::val::undefined();
+        }
     };
     template <typename T>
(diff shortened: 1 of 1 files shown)


finalpatch:
The code change you proposed will trigger the static assertion even when the usage is 100% correct.

bkotsopoulossc:
@finalpatch do you have other suggestions on how best to solve this problem? When you forget to put the right platform flag on a given interface type, and then have code that uses it, because of these fallbacks it ends up failing with a nullptr dereference in very confusing ways. It becomes a big waste of developer time to track these down. What are other approaches would work better?

finalpatch:
If what you want is preventing people from implementing interfaces in languages that the interface is not designated to support implementing (+c +w ...) then we need to generate some extra information.  For example extra enum in interfaces, or a traits class.

You would also make every interface support everything (+c +w +j +o). This would have some code size overhead.  I believe these language flags were there exactly for size reason, to tell the code generator what is not needed.

finalpatch:
This is partially addressed in https://github.com/Snapchat/djinni/pull/26

It's not possible to statically check +c only interfaces at compile time, it has to be a runtime check, and it's covered by the above PR.

We can potentially prevent the use of interfaces that are neither +c nor +w in web assembly,  but that will involve a much bigger change (recursive dependency check at method level), which we might do at a later time.

PR #21, opened 2021-10-20
Add ts-import-name switch

finalpatch:
Add and optional `ts-import-name` switch to allow full customization of the import line.
This defines what string name to use when this TS module is imported by another module through yaml.

eg. if `--ts-module` is "test" and we have 
`--ts-import-name "./test"`

then when this module is imported, the import line will be:
`import {foo} from "./test"`


bkotsopoulossc:
It's up for debate if we want full flexibility or not with a flag. At the very least we need to change this to `import from "./<module_name>"` instead of `import from "<module_name>"` as that doesn't really work out with typical module resolution algorithms (without aliases).

Issue #27, opened 2021-11-12
Cannot build tests at HEAD (c481172185)


jgavris:
Hey folks! I'm very excited that you have taken over `djinni` as we use it extensively at [Butterfly](github.com/ButterflyNetwork/). I have an [internal patch](https://github.com/Snapchat/djinni/pull/28) that isn't in the Dropbox project (after it was archived) and I'd like to share it with the community. I tried running the tests and they are not building for me currently. I'm also super excited you are using Bazel, we have adopted it as well. Is there anything missing from the README or WORKSPACE that would prevent me from running the tests / causing this error?

```sh
jgavris in ~/code/djinni on (HEAD detached at snap/master)
λ git remote -v
origin	git@github.com:ButterflyNetwork/djinni.git (fetch)
origin	git@github.com:ButterflyNetwork/djinni.git (push)
snap	https://github.com/Snapchat/djinni.git (fetch)
snap	https://github.com/Snapchat/djinni.git (push)
jgavris in ~/code/djinni on (HEAD detached at snap/master)
λ git show --stat
commit c48117218536e947d26beb30ddbcab4ac7136eb4 (HEAD, snap/master, record-initializers)
Author: finalpatch <fengli@gmail.com>
Date:   Wed Nov 10 12:19:59 2021 +1100

    Assert with an error message instead of silently return empty object (#26)

    Co-authored-by: Li Feng <li.feng@snapchat.com>

 support-lib/wasm/djinni_wasm.hpp | 10 ++++++++--
 1 file changed, 8 insertions(+), 2 deletions(-)
jgavris in ~/code/djinni on (HEAD detached at snap/master)
λ bazel test //test-suite:all
INFO: Analyzed 14 targets (0 packages loaded, 0 targets configured).
INFO: Found 12 targets and 2 test targets...
ERROR: /Users/jgavris/code/djinni/test-suite/BUILD:153:10: Compiling test-suite/generated-src/wasm/NativeConstantInterfaceWithEnum.cpp failed: (Aborted): wrapped_clang failed: error executing command external/local_config_cc/wrapped_clang '-D_FORTIFY_SOURCE=1' -fstack-protector -fcolor-diagnostics -Wall -Wthread-safety -Wself-assign -fno-omit-frame-pointer -O0 -DDEBUG '-std=c++11' ... (remaining 72 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
[...]

PR #28, opened 2021-11-13
[ObjC] Mark NSObject init / new as unavailable for records with any fields

jgavris:
Blocked by https://github.com/Snapchat/djinni/issues/27 for full test suite, but passes.

```sh
jgavris in ~/code/djinni on objc-record-initializers
λ bazel test //test-suite:djinni-java-tests  //test-suite:djinni-objc-tests
INFO: Analyzed 2 targets (0 packages loaded, 0 targets configured).
INFO: Found 2 test targets...
INFO: Elapsed time: 0.371s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfully, 1 total action
//test-suite:djinni-java-tests                                  (cached) PASSED in 2.7s
//test-suite:djinni-objc-tests                                  (cached) PASSED in 14.9s

Executed 0 out of 2 tests: 2 tests pass.
INFO: Build completed successfully, 1 total action
```

- Records with non-null fields will be improperly constructed if a default initializer (`[NSObject init]` or `[NSObject new]`) is called (on any record with `optional` fields), which can cause an unexpectedly unwrapped optional `nil` in Swift code. Mark these as unavailable to hide them from Swift. 
- Adds `./ci/generate.sh` step as an intermediate check to make sure that all the example test files are 'clean' before running the test suite.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ investigated. Bug reports should have:
 Pull requests are the best way to propose code changes:
 
 1. Fork the repo and create your branch from `master`.
-2. Run `./examples/run_djinni.sh` and check in any changes to generated code.
+2. Run `./ci/generate.sh` and check in any changes to generated code.
 3. Add tests, if appropriate.
 4. If you add or change code generator command line options, update the
    documentation.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,12 +33,13 @@ investigated. Bug reports should have:
 Pull requests are the best way to propose code changes:
 
 1. Fork the repo and create your branch from `master`.
-2. Add tests, if appropriate.
-3. If you add or change code generator command line options, update the
+2. Run `./examples/run_djinni.sh` and check in any changes to generated code.
+3. Add tests, if appropriate.
+4. If you add or change code generator command line options, update the
    documentation.
-4. Ensure the test suite passes with your change.
-5. Test the changes on actual devices if you can.
-6. Issue the pull request.
+5. Ensure the test suite passes with your change.
+6. Test the changes on actual devices if you can.
+7. Issue the pull request.
 
 ## Code style
 
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ either plain Bazel or [Bazelisk](https://github.com/bazelbuild/bazelisk).
 
 ### Building and running the test suite
 
+`./examples/run_djinni.sh` generates the examples sources.
+
 `bazel test //test-suite:all` builds and then runs all tests.
 
 Or `bazel test //test-suite:djinni-objc-tests` and `bazel test
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ either plain Bazel or [Bazelisk](https://github.com/bazelbuild/bazelisk).
 
 ### Building and running the test suite
 
-`./examples/run_djinni.sh` generates the examples sources.
+`./ci/generate.sh` generates the examples sources.
 
 `bazel test //test-suite:all` builds and then runs all tests.
 
--- a/ci/generate.sh
+++ b/ci/generate.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+echo "--- Check Generated files"
+./examples/run_djinni.sh
+./perftest/run_djinni.sh
+./test-suite/run_djinni.sh
+
+if test -z "$(git status --porcelain)"; then
+  echo "--- Success! Generated files are clean."
+else
+  echo "--- Failed! Generated files are dirty after running generator."
+  git status
+  git --no-pager diff --submodule=diff
+  exit 1
+fi
--- a/examples/generated-src/objc/TXSItemList.h
+++ b/examples/generated-src/objc/TXSItemList.h
@@ -4,7 +4,9 @@
 #import <Foundation/Foundation.h>
 
 @interface TXSItemList : NSObject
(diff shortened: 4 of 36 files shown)


finalpatch:
Hi @jgavris thanks for the PR.  I don't completely understand the proposed changes.

We currently do not generate default (no parameter) init method, this makes sure the caller initializes records with proper initial values. 

Are you proposing that we should add a default init method, but hide it from swift if it has nonnull members?  Wouldn't this still expose potential invalid objects to ObjC uses?

examples/generated-src/objc/TXSItemList.h
```diff
@@ -4,7 +4,9 @@
 #import <Foundation/Foundation.h>
 
 @interface TXSItemList : NSObject
-- (nonnull instancetype)initWithItems:(nonnull NSArray<NSString *> *)items;
+- (nonnull instancetype)init NS_UNAVAILABLE;
++ (nonnull instancetype)new NS_UNAVAILABLE;
+- (nonnull instancetype)initWithItems:(nonnull NSArray<NSString *> *)items NS_DESIGNATED_INITIALIZER;
```
jgavris, line 9: @finalpatch The generator does not 'generate' `init` or `new`, but they are inherited from `NSObject`. If you have a record with any non-optional fields, and initialize it say `[TSXItemList new]`, it will throw an implicitly unwrapped optional runtime exception in Swift on accessing `items`. 

Djinni `records` are read-only as well, so it's not possible to set non-optional properties after construction anyway. 

I've also added `NS_DESIGNATED_INITIALIZER` which makes the compiler enforce that any convenience constructor must call this method.

finalpatch: Thanks for the explanation, this make sense.  I didn't know we have always been able to default init records with NSObjct's constructor.

jgavris: No prob! We found this the hard way (in a test case thankfully) but I figured it was a quick and easy patch and modernizes the generated Obj-C code (especially for Swift consumption).

finalpatch:
My main concern is that this is a breaking change and it may break some existing code.  However,  if I understand correctly, if any code is broken by this change,  then that code itself is problematic and should be fixed.

jgavris:
There will likely be code that is 'accidentally' using the `NSObject` constructors (ex; `TSXItemList()` in Swift) and will have to migrate to the designated initializer, but they are not safe usages currently, no.

CONTRIBUTING.md
```diff
@@ -33,12 +33,13 @@ investigated. Bug reports should have:
 Pull requests are the best way to propose code changes:
 
 1. Fork the repo and create your branch from `master`.
-2. Add tests, if appropriate.
-3. If you add or change code generator command line options, update the
+2. Run `./examples/run_djinni.sh` and check in any changes to generated code.
```
finalpatch, line 36: If you modified the code generator, run `run_djinni.sh` in `./examples`, `./test-suite`, `./perftest` and check in any changes to generated code.

README.md
```diff
@@ -19,6 +19,8 @@ either plain Bazel or [Bazelisk](https://github.com/bazelbuild/bazelisk).
 
 ### Building and running the test suite
 
+`./examples/run_djinni.sh` generates the examples sources.
```
finalpatch, line 22: There are also `run_djinni.sh` scripts in other directories.  Maybe let's just say "use the run_djinni.sh scripts to generate bridge code"

finalpatch (approved):
looks good once you re-run codegen in the other directories.

finalpatch:
@jgavris do you still intend to merge this PR?  Like I said above,  we need to re-run codegen in all sub-project directories and fix the documentation otherwise I think this is good to go.

jgavris:
@finalpatch Yes, thanks for the feedback. I was going to suggest adding your comment in https://github.com/Snapchat/djinni/pull/28#discussion_r748942759 to the CI workflow (and test step that ensures they are not dirty), so that future contributions are 'clean'.

finalpatch:
> @finalpatch Yes, thanks for the feedback. I was going to suggest adding your comment in [#28 (comment)](https://github.com/Snapchat/djinni/pull/28#discussion_r748942759) to the CI workflow (and test step that ensures they are not dirty), so that future contributions are 'clean'.

Thanks.  Yeah it's a bit annoying we have to check in the codegen changes.  I'm not sure if the CI workflow is the right place for it.  We are exploring way to run the codegen as a part of the bazel build process,  hopefully this will make the manual codegen step no longer necessary.

ci/generate.sh
```diff
@@ -0,0 +1,17 @@
...
+
+set -euo pipefail
+
+echo "--- Check Examples files"
+./examples/run_djinni.sh
+./perftest/run_djinni.sh
+./test-suite/run_djinni.sh
+
+if test -z "$(git status --porcelain)"; then
+  echo "--- Success! Examples files are clean."
```
finalpatch, line 11: Can we change "Example files" to "generated files"?

jgavris: 👌

CONTRIBUTING.md
jgavris: Done

README.md
jgavris: Done

jgavris:
>We are exploring way to run the codegen as a part of the bazel build process, hopefully this will make the manual codegen step no longer necessary.

SGTM. It's a bit tricky to get Bazel to modify / generate files as part of the build https://shantanugoel.com/2020/05/03/bazel-rule-auto-generate-files-compile-time/

PR #31, opened 2021-12-03
Experimental support for async interface via the future<> generic type

finalpatch:
This PR adds support for a `future<>` type to the djinni idl.  The `future<>` djinni type is mapped to the `Future` type defined in the C++, Java and ObjC djinni support library (also added in this PR).  In Javascript, `future<>` is mapped to the builtin `Promise` type (and therefore supports the JS await syntax).

Basically, instead of writing this:
```
FooCb = interface {
  complete(res: i32): void;
}
Foo = interface {
  bar(cb: FooCb): void;
}
```

We can now express this more succinctly with :
```
Foo = interface {
  bar(): future<i32>;
}
```

The "Future" objects are fulfilled by calling the `setValue()` method on their associated  `Promise` object (note this is not related to the above mentioned JS Promise type) in C++, Java and ObjC.  The Promise/Future API closely resemble the std::experimental::promise/future types defined by the C++ concurrency TS. 

The main benefit of this feature is that we no longer need to define new callback interfaces just to return a value asynchronously.  The `future<>` objects as implemented by this PR can be passed across language boundaries and back, which can be utilised to support cross language asynchronous workflows.

If C++ coroutine support is available (with C++20 or with `-fcoroutine-ts`), then the C++ Future<> class can be used as an awaitable object. Making it easier to consume asynchronous results.

This PR includes a djinni code generator [change borrowed from the cross-language-cpp repo](https://github.com/cross-language-cpp/djinni-generator/pull/113) to support generic external types in ObjC,


support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,193 @@
...
+template <typename T>
+class Future;
+
+template <typename T>
+class PromiseBase {
+public:
+    Future<T> getFuture();
+
+protected:
+    void setValue(typename ValueHolder<T>::type val) {
```
bholmes-sc, line 64: It's worth adding comment here about how this changes the internal state and the downstream effect on the handler

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,193 @@
...
+    T getValueUnsafe() const {return *value;}
+};
+template <>
+struct ValueHolder<void> {
+    using type = bool;
+    std::optional<bool> value;
+    void getValueUnsafe() const {}
+};
+
+template<typename T>
```
bholmes-sc, line 40: Let's add a comment describing this class and it's basic behavior.

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,193 @@
...
+            std::unique_lock lk(_sharedState->mutex);
+            _sharedState->exception = ex;
+            sharedState = std::move(_sharedState);
+        }
+        callResultHandler(sharedState);
+    }
+private:
+    SharedStatePtr<T> _sharedState = std::make_shared<SharedState<T>>();
+
+    void callResultHandler(SharedStatePtr<T> sharedState) {
```
bholmes-sc, line 91: Can we mark this static? There's no reference to member vars.

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,193 @@
...
+        }
+        callResultHandler(sharedState);
+    }
+
+    template <typename E>
+    void setException(const E& ex) {
+        setException(std::make_exception_ptr(ex));
+    }
+
+    void setException(std::exception_ptr ex) {
```
bholmes-sc, line 79: If calling both setValue and setException is a violation then let's call that out in the coments.  We can just state that the internal state of the class is invalid once called.

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,193 @@
...
+template <typename T>
+class PromiseBase {
+public:
+    Future<T> getFuture();
+
+protected:
+    void setValue(typename ValueHolder<T>::type val) {
+        SharedStatePtr<T> sharedState;
+        {
+            std::unique_lock lk(_sharedState->mutex);
```
bholmes-sc, line 67: This will crash on null pointer deref after setValue() is called right?   Is this behavior we should try to explicitly catch?

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,193 @@
...
+        if (sharedState->isReady()) {
+            // result already available
+            continuation(sharedState);
+        } else {
+            // result not yet available
+            sharedState->handler = std::move(continuation);
+        }
+        return nextFuture;
+    }
+private:
```
bholmes-sc, line 184:
```suggestion

private:
```

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,193 @@
...
+        return _sharedState != nullptr;
+    }
+    
+    bool isReady() const {
+        std::unique_lock lk(_sharedState->mutex);
+        return _sharedState->isReady();
+    }
+    
+    void wait() const {
+        std::unique_lock lk(_sharedState->mutex);
```
bholmes-sc, line 138: so, this locks the object until wait() is complete.   if someone calls isReady() while another thread is calling wait(), won't this cause isReady to hang until wait() is complete?

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,193 @@
...
+        auto nextFuture = nextPromise.getFuture();
+        auto continuation = [this, handler, nextPromise] (SharedStatePtr<T> x) mutable {
+            try {
+                if constexpr(std::is_void_v<HandlerReturnType>) {
+                    handler(Future<T>(x));
+                    nextPromise.setValue();
+                } else {
+                    nextPromise.setValue(handler(Future<T>(x)));
+                }
+            } catch (std::exception& e) {
```
bholmes-sc, line 171: can we disable the try/catch in the event -fno-exceptions is set?

support-lib/java/com/snapchat/djinni/Future.java
```diff
@@ -0,0 +1,107 @@
...
+        sharedState = state;
+    }
+
+    public boolean isReady() {
+        synchronized(sharedState) {
+            return sharedState.isReady();
+        }
+    }
+
+    public T get() throws Throwable {
```
bholmes-sc, line 41: does need to be declared as throwing?

support-lib/jni/Future_jni.cpp
```diff
@@ -0,0 +1,38 @@
...
+  * limitations under the License.
+  */
+
+#include "Future_jni.hpp"
+
+namespace snapchat {
+namespace djinni {
+
+// NOLINTNEXTLINE
+static void NativeFutureHandler_nativeHandleResult(JNIEnv* jniEnv, jclass /*unused*/, jlong nativeFunc, jlong nativePromise, jobject jres, jthrowable jex) {
```
bholmes-sc, line 23:
```suggestion
static void NativeFutureHandler_nativeHandleResult(
        JNIEnv* jniEnv, jclass /*unused*/, jlong nativeFunc, jlong nativePromise, jobject jres, jthrowable jex) {
```

support-lib/jni/Future_jni.cpp
```diff
@@ -0,0 +1,38 @@
...
+  * distributed under the License is distributed on an "AS IS" BASIS,
+  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
+  */
+
+#include "Future_jni.hpp"
+
+namespace snapchat {
+namespace djinni {
```
bholmes-sc, line 20:
```suggestion
namespace snapchat::djinni {
```

support-lib/jni/Future_jni.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+#pragma once
+
+#include "djinni_support.hpp"
+#include "Marshal.hpp"
+#include "../cpp/Future.hpp"
+
+namespace snapchat {
+namespace djinni {
+struct PromiseJniInfo {
```
bholmes-sc, line 25:
```suggestion

struct PromiseJniInfo {
```

support-lib/jni/Future_jni.hpp
```diff
@@ -0,0 +1,136 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "djinni_support.hpp"
+#include "Marshal.hpp"
+#include "../cpp/Future.hpp"
+
+namespace snapchat {
```
bholmes-sc, line 23: We should eventually move all of these classes into the `snap`. namespace

test-suite/djinni/test.djinni
```diff
@@ -41,6 +47,11 @@ test_helpers = interface +c {
     static assorted_primitives_id(i: assorted_primitives): assorted_primitives;
 
     static id_binary(b: binary): binary;
+
+    static get_async_result(): future<i32>;
+    static future_roundtrip(f: future<i32>): future<string>;
```
bholmes-sc, line 52: add a comment for this one

test-suite/djinni/test.djinni
```diff
@@ -1,3 +1,9 @@
+@extern "../../support-lib/future.yaml"
+
+async_interface = interface +j +o +w {
```
bholmes-sc, line 3: why snake case here for async_interface?  These are normally Pascal Case, eg AsyncInterface

test-suite/handwritten-src/cpp/test_helpers.cpp
```diff
@@ -161,4 +168,42 @@ std::vector<uint8_t> TestHelpers::id_binary(const std::vector<uint8_t> & v) {
...
+    return f;
+}
+
+snapchat::djinni::Future<std::string> TestHelpers::future_roundtrip(snapchat::djinni::Future<int32_t> f) {
+    return f.then([] (snapchat::djinni::Future<int32_t> f) {
+        return std::to_string(f.get());
+    });
+}
+
+snapchat::djinni::Future<std::string> TestHelpers::check_async_interface(const std::shared_ptr<AsyncInterface> & i) {
```
bholmes-sc, line 198: Are there any C++ unit tests calling this?

I don't see any pure C++ unit tests.  I also don't see an tests on calls to setException, but might have just missed this one.

bholmes-sc (approved):
Really impressive!

test-suite/djinni/test.djinni
finalpatch: all the other djinni test interfaces are snake_case

test-suite/handwritten-src/cpp/test_helpers.cpp
finalpatch: there is a `testFutureRoundtripWithException` test that throws an exception in platform code which will indirectly `setException`

test-suite/djinni/test.djinni
bholmes-sc: my question is oriented at the general case

bholmes-sc:
lgtm

support-lib/cpp/Future.hpp
finalpatch: no.  the mutex is release while a condition variable is waiting.

support-lib/cpp/Future.hpp
finalpatch: Let's do this later.  Right now there are so many try/catches in djinni and its generated code there's no point to attempt no-exceptions here.

support-lib/java/com/snapchat/djinni/Future.java
finalpatch: calling `get()` will throw the stored exception.

test-suite/djinni/test.djinni
finalpatch: The convention is use snake_case in djinni idl files.  The generated code style is defined by the command line, eg `--ident-jni-class NativeFooBar`.

bholmes-sc:
lgtm

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,307 @@
...
+        if (sharedState) {
+            return Future<T>(sharedState).get();
+        } else {
+            return Future<T>(_coroState).get();
+        }
+    }
+    bool await_suspend(std::experimental::coroutine_handle<> h) {
+        this->then([h, this] (Future<T> x) mutable {
+            _coroState = x._sharedState;
+            h();
```
finalpatch, line 293: H.destroy();

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,307 @@
...
+        auto nextFuture = nextPromise->getFuture();
+        auto continuation = [handler = std::forward<FUNC>(handler), nextPromise = std::move(nextPromise)] (detail::SharedStatePtr<T> x) mutable {
+            try {
+                if constexpr(std::is_void_v<HandlerReturnType>) {
+                    handler(Future<T>(x));
+                    nextPromise->setValue();
+                } else {
+                    nextPromise->setValue(handler(Future<T>(x)));
+                }
+            } catch (std::exception& e) {
```
dchapp-sc, line 253: Can we make this a const reference? If only to signal that nothing will modify `e`.

support-lib/wasm/Future_wasm.hpp
```diff
@@ -0,0 +1,109 @@
...
+    private:
+        em::val _resolveFunc = em::val::undefined();
+        em::val _rejectFunc = em::val::undefined();
+        std::optional<Future<CppResType>> _future;
+
+        // runs in main thread
+        void doResolve() {
+            try {
+                _resolveFunc(RESULT::Boxed::fromCpp(_future->get()));
+            } catch (std::exception& e) {
```
dchapp-sc, line 82: Can be const ref?

support-lib/jni/Future_jni.hpp
```diff
@@ -0,0 +1,136 @@
...
+        promise->set(jniEnv, jniEnv->NewObject(promiseJniInfo.clazz.get(), promiseJniInfo.constructor));
+        auto future = ::djinni::LocalRef<jobject>(jniEnv, jniEnv->CallObjectMethod(promise->get(), promiseJniInfo.method_get_future));
+        ::djinni::jniExceptionCheck(jniEnv);
+
+        c.then([promise, &promiseJniInfo] (Future<CppResType> cppFuture) {
+            JNIEnv* jniEnv = ::djinni::jniGetThreadEnv();
+            try {
+                auto res = cppFuture.get();
+                jniEnv->CallVoidMethod(promise->get(), promiseJniInfo.method_set_value, RESULT::Boxed::fromCpp(jniEnv, res).get());
+            } catch (std::exception& e) {
```
dchapp-sc, line 123: Can be const ref?

support-lib/objc/Future_objc.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+    static ObjcType fromCpp(CppType c)
+    {
+        DJPromise<typename RESULT::Boxed::ObjcType>* promise = [[DJPromise alloc] init];
+        DJFuture<typename RESULT::Boxed::ObjcType>* future = [promise getFuture];
+
+        c.then([promise] (Future<CppResType> res) {
+                try {
+                    [promise setValue:RESULT::Boxed::fromCpp(res.get())];
+                } catch (std::exception& e) {
```
dchapp-sc, line 63: Can be const ref?

dchapp-sc (changes requested):
Overall LGTM, modulo a few nitpicks.

support-lib/cpp/Future.hpp
finalpatch: actually, no.  our coroutine support library returns `std::suspend_never` in its `final_suspend()` method. so the handle is automatically destroyed when the coroutine returns.

test-suite/BUILD
```diff
@@ -54,6 +54,7 @@ objc_library(
     name = "djinni-tests-objcxx",
     copts = [
         "-ObjC++",
+        "-std=c++17",
```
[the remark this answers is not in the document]
finalpatch: We have `build --cxxopt=-std=c++17`  in `.bazelrc`
Have to add this line here because `--cxxopt` is not automatically added to objc files.

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,317 @@
...
+    std::unique_ptr<ValueHandlerBase<T>> handler;
+
+    bool isReady() const {
+        return this->value.has_value() || exception != nullptr;
+    }
+};
+
+template<typename T>
+using SharedStatePtr = std::shared_ptr<SharedState<T>>;
+}
```
bholmes-sc, line 99:
```suggestion

} // namespace detail
```

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+template <typename T>
+class Future;
```
bholmes-sc, line 102: are we ending the detail namespace on line 99 and starting again after this, just for the forward decl here?  can we just declare this above line 41, or should this be part of another header?

support-lib/cpp/Future.hpp
```diff
@@ -0,0 +1,317 @@
...
+        if (handler) {
+            // handler already assigned, call it inline
+            handler->call(sharedState);
+        } else {
+            // otherwise unblock potential waiters
+            sharedState->cv.notify_all();
+        }
+    }
+};
+}
```
bholmes-sc, line 160:
```suggestion

} // namespace detail
```

dchapp-sc (approved):
LGTM

support-lib/objc/DJFuture.h
```diff
@@ -0,0 +1,45 @@
...
+  */
+
+#pragma once
+
+#import <Foundation/Foundation.h>
+
+@class DJFuture;
+
+@interface DJFuture<__covariant Value> : NSObject
+-(nonnull instancetype) init NS_UNAVAILABLE;
```
chanryu, line 24:
```suggestion
- (nonnull instancetype)init NS_UNAVAILABLE;
```
- Space after the leading `-` (or `+`)
- No space before the method name `init`.

support-lib/objc/DJFuture.h
```diff
@@ -0,0 +1,45 @@
...
+// If the future is ready, then calling its `get` method will not block.
+-(BOOL) isReady;
+// Block and wait for the result (or exception). This can only be called once.
+-(nullable Value) get;
+// Tell the future to Call the specified handler routine when it becomes
+// ready. Returns a new future that wraps the return value of the handler
+// routine. The current future becomes invalid after this call.
+-(nonnull DJFuture<id>*)then:(_Nullable id(^_Nonnull)(DJFuture<Value>* _Nonnull))handler;
```
chanryu, line 32:
```suggestion
- (nonnull DJFuture<id> *)then:(_Nullable id(^_Nonnull)(DJFuture<Value> * _Nonnull))handler;
```
- Space before `*` (like C)

support-lib/objc/DJFuture.h
```diff
@@ -0,0 +1,45 @@
...
+
+// ------------------------------------------
+
+@interface DJPromise<Value> : NSObject
+-(nonnull instancetype) init;
+// Get a future object associated with this promise
+-(nonnull DJFuture<Value>*) getFuture;
+// `setValue` or `setException` can only be called once on a promise. After
+// which the internal state becomes invalid.
+-(void) setValue:(nullable Value) val;
```
chanryu, line 43:
```suggestion
-(void) setValue:(nullable Value)val;
```

- No space before the arg `val`.

support-lib/objc/DJFuture.mm
```diff
@@ -0,0 +1,170 @@
...
+
+typedef _Nullable id (^Continuation)(DJSharedSate* _Nonnull);
+
+@interface DJSharedSate<Value> : NSObject
+    @property (nonatomic, strong) Value value;
+    @property (nonatomic, strong) NSException* exception;
+    @property (nonatomic, strong) NSCondition* cond;
+    @property (nonatomic, strong) Continuation handler;
+
+    @property (readonly) BOOL isReady;
```
chanryu, line 29:
```suggestion
@property (nonatomic, strong) Value value;
@property (nonatomic, strong) NSException *exception;
@property (nonatomic, strong) NSCondition *cond;
@property (nonatomic, strong) Continuation handler;

@property (readonly) BOOL isReady;
```
- No indent
- `strong` is not required but ok.

support-lib/objc/DJFuture.mm
```diff
@@ -0,0 +1,170 @@
...
+@end
+
+@implementation DJSharedSate
+-(instancetype)init {
+    if (self = [super init]) {
+        self->_value = nil;
+        self->_exception = nil;
+        self->_cond = [[NSCondition alloc] init];
+        self->_handler = nil;
```
chanryu, line 38:
```suggestion
        _cond = [[NSCondition alloc] init];
```

- `self->` not necessary
- all ivars are initialized to zeros by default. so, no need to initialized to `nil`.

support-lib/objc/DJFuture.mm
```diff
@@ -0,0 +1,170 @@
...
+@end
+
+// ------------------------------------------
+
+@implementation DJPromise {
+    DJSharedSate* _sharedState;
+}
+-(instancetype) init {
+    if (self = [super init]) {
+        _sharedState =[[DJSharedSate alloc] init];
```
chanryu, line 132:
```suggestion
        _sharedState = [[DJSharedSate alloc] init];
```

test-suite/handwritten-src/objc/tests/DBAsyncTest.m
```diff
@@ -0,0 +1,79 @@
...
+    DJFuture<NSString*>* f3 = [DBTestHelpers futureRoundtrip:f2];
+    [p setValue:@"36"];
+    XCTAssertEqualObjects([f3 get], @"36");
+}
+
+- (void) testFutureRoundtripWithException {
+    DJPromise<NSString*>* p = [[DJPromise alloc] init];
+    DJFuture<NSString*>* f = [p getFuture];
+    DJFuture<NSNumber*>* f2 = [f then:^id(DJFuture<NSString*>* s) {
+            if (1) {
```
chanryu, line 58: `if (1)` ?

chanryu (commented):
Just some comments on the indentation style -- based on the generally accepted rules.

test-suite/handwritten-src/objc/tests/DBAsyncTest.m
```diff
@@ -0,0 +1,79 @@
...
+- (void)testConsumeNativeFuture {
+    DJFuture<NSNumber*>* f1 = [DBTestHelpers getAsyncResult];
+    DJFuture<NSString*>* f2 =[f1 then:^id(DJFuture<NSNumber*>* res) {
+            return [[res get] stringValue];
+        }];
+    DJFuture<NSNumber*>* f3 = [f2 then:^id(DJFuture<NSString*>* res) {
+            return @([[res get] integerValue]);
+        }];
+    NSNumber* i = [f3 get];
+    XCTAssertEqual([i integerValue], 42);
```
chanryu, line 40: It took me 5 minutes for me to figure out what's with this magic number `42`.
It would be nice to see it in this test. e.g.,
```ObjC
DJFuture<NSNumber *> *f1 = [DBTestHelpers getAsyncResultWithValue:@42];
```

test-suite/handwritten-src/objc/tests/DBAsyncTest.m
```diff
@@ -0,0 +1,79 @@
...
+        [f3 get];
+    } @catch (NSException* e) {
+        s = e.reason;
+    }
+    XCTAssertEqualObjects(s, @"123");
+}
+
+- (void) testFutureRoundtripBackwards {
+    DJFuture<NSString*>* s = [DBTestHelpers checkAsyncInterface: [[AsyncInterfaceImpl alloc] init]];
+    XCTAssertEqualObjects([s get], @"36");
```
chanryu, line 76: Same as `42`.

test-suite/handwritten-src/objc/tests/DBAsyncTest.m
finalpatch: Yes I want to always throw in this function, but without this "if", the compile would complain about either return type or dead code.

test-suite/handwritten-src/objc/tests/DBAsyncTest.m
finalpatch: good idea

finalpatch: I tried that but it's not straightforward because of this code:
```
#if defined(__EMSCRIPTEN__)
    emscripten_async_call([] (void* context) {
        auto* p = reinterpret_cast<snapchat::djinni::Promise<int32_t>*>(context);
        p->setValue(42);
        delete p;
    }, p, 10/*ms*/);
#else
```
Here `emscripten_async_call` must take a function pointer.  If I capture anything into this lambda here, it will not be convertible to a plain function pointer.

chanryu: I'm not sure if I'm really following you on this. Can't we just do this?

```cpp
snapchat::djinni::Future<int32_t> TestHelpers::get_async_result(int32_t value) {
     // ...

 #if defined(__EMSCRIPTEN__)
     emscripten_async_call([value] (void* context) {
         auto* p = reinterpret_cast<snapchat::djinni::Promise<int32_t>*>(context);
         p->setValue(value);
         delete p;
     }, p, 10/*ms*/);
 #else
```

test-suite/handwritten-src/objc/tests/DBAsyncTest.m
chanryu: It would be nice to have the above as a comment in the code.
This test file is also a showcase of this new addition to djinni.

test-suite/handwritten-src/objc/tests/DBAsyncTest.m
finalpatch: No, this will not work.  A lambda like this `[] { ... }` (without captures) can be converted to a plain C function pointer.  While a lambda like this `[value] { ... }` cannot be converted to a C function pointer (it is its own special lambda type).  The function `emscripten_async_call` expects a function pointer,  passing it a lambda will produce a compilation error.

We can, of course, allocate a context struct on the heap and put the value there, but that is a  bit too much trouble than its worth.

chanryu: Fair enough!

chanryu (approved):
Just need some more indentation fixes in DBAsyncTest.m, especially for the position of `*`.
Otherwise, LGTM.

PR #33, opened 2021-12-14
Fix JNI enum flags in yaml output

finalpatch:
Fix the yaml output for enum flags.  flags should produce "Ljava/util/EnumSet;" rather than "LPackage/Class;" in yaml as well as in local codegen.

The bug was found by @delaitre-manfrotto  and was fixed in the cross-language-cpp repo on the yaml consumer side (https://github.com/cross-language-cpp/djinni-generator/pull/90). After examine this problem closely, I believe this is better fixed on the yaml producer side, as the `typeSignature` field in the yaml  file is incorrect in the first place.

Testing: examined `yaml-test.yaml` in `test-suite/djinni-output-temp/yaml`,  and the `typeSignature` fields now have the correct value `Ljava/util/EnumSet;`


delaitre-manfrotto:
Seems like a better fix than my original one indeed.

bholmes-sc (approved):

PR #35, opened 2021-12-15
Fix js exception handling

bkotsopoulossc:
When using exception handling in emscripten, the framework does some tricks to get around the lack of native support in wasm (coming soon). Essentially what it does is takes every c++ function that needs a try-catch around it, and generates a JS wrapper, and then instead of calling the c++ function directly, traps out of wasm to JS, and JS calls the c++ function. The JS wrapper than has some logic in the `catch` block, that essentially says, if it's a JS exception, just re-throw it, and if it's a c++ one, just mark it as thrown with a global flag, but do nothing. 

What this means is, we expect the following:
- JS calls into C++, C++ throws. JS catches an exception which is just a number, and we have to go back to C++ to cast it to a pointer, and read the message
- JS calls into C++, C++ calls a JS object, JS throws. The JS error should bubble all the way back up to the top-level JS caller, and it should not be a number. It should be the untouched JS Error object.

There currently is a wrapper in the support lib, where the JS proxy objects are, which catches JS exceptions, and re-throws as a c++ one, with the message copied over.

This throws away useful information, because the JS error has things like a type, a name, a stack trace, etc. Instead, we want to keep emscripten's designed behaviour, and let that JS error bubble up. That way, when we catch it several levels up, we should have the full stack trace.

It looks like the readme and scripts are either broken or undocumented. I tried to run the tests with `bzl run //test-suite:server-ts` or `bzl test //test-suite:server-ts` but neither actually ran the tests. I've tested this on web, and written out some unit tests that roughly do what I want, but not sure how to run them. It would be great to have CI take care of all of this...


finalpatch:
@bkotsopoulossc  `bzl run //test-suite:server-ts` should work if you have all the dependencies installed.  This is the command I use to test this stuff.

finalpatch:
Does this mean if C++ code calls JS and JS code throws an exception, there is no way to handle it in C++ code. The only way to handle JS exception is by another layer of JS code above the C++ code?

test-suite/handwritten-src/js/CppExceptionTest.js
```diff
@@ -8,15 +8,26 @@ class CppExceptionTest {
     }
 
     testCppException() {
-        var thrown = null;
         try {
             this.cppInterface.throwAnException();
+            fail();
         } catch (e) {
-            thrown = e;
+            assertNe(thrown, null);
```
finalpatch, line 15: use undefined var?

bkotsopoulossc:
This is what I get when running the tests. It builds the wasm binary but I'm not sure what else I expect to happen. I'm not sure what setup stuff I'm supposed to do beforehand, but can we do it in bazel automatically? Or document it?
```
% bzl run //test-suite:server-ts
2021-12-15 17:56:58 EST Running native bzl
INFO: Invocation ID: 5c1e3a39-d7a5-4248-9615-02b0f3fecafe
INFO: Analyzed target //test-suite:server-ts (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //test-suite:server-ts up-to-date:
  bazel-bin/test-suite/server-ts
INFO: Elapsed time: 0.387s (0.01m), Critical Path: 0.00s (0.00m)
INFO: 1 process: 1 internal (0.00%).
INFO: Running command line: bazel-bin/test-suite/server-ts
INFO: Build Event Protocol files produced successfully.
INFO: Build completed successfully, 1 total action

                                                                               
                This is not the tsc command you are looking for                
                                                                               

To get access to the TypeScript compiler, tsc, from the command line either:

- Use npm install typescript to first add TypeScript to your project before using npx
- Use yarn to avoid accidentally running code from packages which haven't been installed
/Users/brad.kotsopoulos/Snapchat/Dev/.cache/_bazel_root/40029df34a6f5ca60d83b822d09254a4/execroot/snap_djinni/bazel-out/darwin-fastbuild/bin/test-suite/server-ts: line 5: browserify: command not found
```

test-suite/handwritten-src/js/CppExceptionTest.js
bkotsopoulossc: I'm sure once I can run tests I'll be able to sort these things out. Also not sure if there is a ton of value anyway in maintaining the javascript ones if we have typescript

finalpatch:
After some testing, I find the whole cross language exception thing is rather broken in the current version of emscripten.

* If I throw a non-int value from JS then
   * C++ code cannot catch it even with catch(...).
   * C++ stack unwinding is not performed in this case.
   * The exception can be caught by an upstream JS catch clause.
* If I throw an int value from JS
   * C++ code can catch it with catch(...)
   * C++ stack unwinding is performed
   * Upstream JS catch clause does not see this exception (consumed in C++, as expected)
   * Curiously, the actual exception object caught in C++ is the previous exception object thrown by C++ code.
* If I throw an int value from JS and do not catch it in C++
   * C++ stack unwinding is performed
   * The exception can be caught by an upstream JS catch clause.

bkotsopoulossc:
Thanks for the analysis here. Not being able to catch in c++ doesn't worry me, but not unwinding the stack and calling the destructors won't work. I'll close this and put up a simpler proposal.

I wonder what results we'd see if we did the same level of analysis on exceptions thrown from c++. Emscripten does a lot of magic to make exceptions work. I wouldn't be surprised if some of the cases don't work correctly with c++ exceptions either (ex. c++ throws, it propagates through a mix of JS and C++ frames, any frame can catch and/or re-throw, does stack unwinding remain consistent all the way up)

support-lib/wasm/djinni_wasm.cpp
```diff
@@ -162,16 +161,15 @@ EM_JS(void, djinni_init_wasm, (), {
         };
 
         Module.callJsProxyMethod = function(obj, method, ...args) {
-            try {
-                return obj[method].apply(obj, args);
-            } catch (e) {
-                return e;
```
bkotsopoulossc, line 168: This is also somewhat dangerous @LiFengSC because we're catching everything, but only re-throwing in c++ if it's an instanceof Error. If anything else gets thrown, we're returning it as if it was the actual return type, which it wasn't . This seems scary if there are ever cases that throw non-Error types

finalpatch: Right.  We should check the type of `e` in JS, if it's not an `Error`, then wrap it in an `Error` first.

PR #37, opened 2021-12-17
Propagate stack from JS Error into C++ exception

bkotsopoulossc:
Pivot from https://github.com/Snapchat/djinni/pull/35/.

This solves the problem of information loss, by taking all of the useful info out of the JS error before turning it into a c++ exception.

### DIFF
--- a/support-lib/wasm/djinni_wasm.cpp
+++ b/support-lib/wasm/djinni_wasm.cpp
@@ -60,7 +60,20 @@ const em::val& JsProxyBase::_jsRef() const {
 
 void JsProxyBase::checkError(const em::val& v) {
     if (v.instanceof(em::val::global("Error"))) {
-        throw JsException(v["message"].as<std::string>());
+         // The stack property is non-standard, but well supported in browsers
+         // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Stack#browser_compatibility
+         // It also seems to include the name and message properties, so no need to access them separately
+         //
+         // >>> function doThrow() { throw new Error("foo")}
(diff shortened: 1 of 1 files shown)


support-lib/wasm/djinni_wasm.cpp
```diff
@@ -60,7 +60,20 @@ const em::val& JsProxyBase::_jsRef() const {
...
+         // >>> function doThrow() { throw new Error("foo")}
+         // >>> try { doThrow() } catch (e) { console.log(e.stack) }
+         // Error: foo
+         //     at doThrow (<anonymous>:1:25)
+         //     at <anonymous>:1:7
+         // >>> try { doThrow() } catch (e) { console.log(e.name) }
+         // Error
+         // >>> try { doThrow() } catch (e) { console.log(e.message) }
+         // foo
+        throw JsException(v["stack"].as<std::string>());
```
finalpatch, line 76: should we capture both `stack` and `message`?

bkotsopoulossc: Nope. If you look at the comment I left (L65) plus the examples, it includes the message in the field as it is.

finalpatch (approved):
I wonder if we could use `stack` when it's available, but fallback to `message` if it's not.

bkotsopoulossc:
We can, but if you look at browser compatibility table I put in the comments as well, we’re not going to find a browser that supports wasm but not stack

finalpatch:
> We can, but if you look at browser compatibility table I put in the comments as well, we’re not going to find a browser that supports wasm but not stack

sure. i just feel that makes the code not just correct in practice, but also in principle.

PR #40, opened 2021-12-21
Refactors

finalpatch:
A bunch of refactors to get the repo ready for release packages (making it easier to consume the support lib from any path in user's source tree)

* Move all C++ include files under `cpp`
* Move all classes from `::snapchat::djinni` namespace to `::djinni`
* Remove the codegen level support for the `outcome` type, and make it a yaml type
* Move the `outcome` marshaller code to it own files `Outcome_*.hpp`
* Add a codegen option to specify C++ support library include path `cpp-base-lib-include-prefix`
* Allow using the `$` symbol in yaml header paths to represent the "base lib include prefix"


src/source/YamlGenerator.scala
```diff
@@ -148,7 +148,7 @@ class YamlGenerator(spec: Spec) extends Generator(spec) {
       "header" -> QuotedString(objcMarshal.include(td.ident)),
       "boxed" -> QuotedString(objcMarshal.boxedTypename(td)),
       "pointer" -> objcMarshal.isPointer(td),
-      "generic" -> false,
+      // "generic" -> false,
```
bholmes-sc, line 151: what is this change for?  

should this commented out code just be removed?

finalpatch: Djinni generates yaml for each user defined type it processes. Since we can't define generic interfaces in the IDL itself, the "generic" field being true or false does not really matter for IDL defined types.  This field is only useful in external types, which the user will provide custom marshallers. 

Historically Djinni writes "generic" -> true in the JNI section, and nothing in the other sections.  Since we have added generic support for ObjC, we do have this field in objc types, so I wrote it out in the yaml too.  But writing it out explicitly causes a lot of the existing yaml files to be modified even though the new field merely has the default value,  which does not impact codegen at all.

Here I choose to omit the field in the yaml generator (and in the case of JNI, keep it true) when it does not impact codegen, to minimize the generated code change.

paulo-coutinho:
There are a lot of comments on code. It can be removed? Because no one will remember why it is commented there and can be confusing.

finalpatch:
> There are a lot of comments on code. It can be removed? Because no one will remember why it is commented there and can be confusing.

@paulo-coutinho you are right, I forgot to delete them.

bholmes-sc (approved):

PR #41, opened 2021-12-23
Fix race condition when finalizing weakrefs

finalpatch:
When a C++ object's proxy is expired in JS but the finalizer is not yet run,  the same object gets a new proxy when it's passed across the boundary.  Because both proxies are registered with the finalizer, the finalizer `nativeDestroy` will be called twice, and the second time it'll trigger an assertion, as the proxy is no longer in the proxy cache.

The sequence of events are like this:

1. C++ object proxy in JS is expired and becomes garbage
2. The same C++ is passed to JS again
3. The marshalling code sees the weakref to the original proxy is expired, and creates a new proxy
4. The expired proxy object's finalizer is called, and removes the new proxy from proxy cache
5. The new proxy object is expired in JS
6. The new proxy object's finalizer is called, but it can't find itself in the proxy cache and triggers assertion

This PR fixes the problem by keeping a count in the proxy cache.  When an object gets a new proxy but the old proxy's finalizer is not yet run,  we'll bump the count.  We'll only remove the proxy cache entry when the count reaches 0.

### DIFF
--- a/support-lib/wasm/djinni_wasm.cpp
+++ b/support-lib/wasm/djinni_wasm.cpp
@@ -42,7 +42,7 @@ Date::JsType Date::fromCpp(const CppType& c) {
 
 JsProxyId nextId = 0;
 std::unordered_map<JsProxyId, std::weak_ptr<JsProxyBase>> jsProxyCache;
-std::unordered_map<void*, em::val> cppProxyCache;
+std::unordered_map<void*, CppProxyCacheEntry> cppProxyCache;
 std::mutex jsProxyCacheMutex;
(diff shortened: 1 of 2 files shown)


paulo-coutinho:
Make sense use some specific type for `counter` variable instead of `int`?

Example: https://github.com/waTeim/poco/blob/6399b13a398f0f437e2b04ff8d49d2f5c27eb7aa/include/Poco/AtomicCounter.h#L127-L139

finalpatch:
No need for atomic because it's already protected by a mutex

bholmes-sc (approved):

PR #42, opened 2021-12-23
add support for executable

paulo-coutinho:


paulo-coutinho:
Add support to execute generator from bazel:

```
bazel run //generator:generator
```

PR #45, opened 2022-01-04
Support version

paulo-coutinho:
Hi,

It add support for version number in Java JAR and add parameter "--version" to work.

I have a problem with scala_binary that don't have support for manifest lines.

There is an open issue:
bazelbuild/rules_scala#166

Only merge after it be solved. But i let it open to team check the code.

Thanks.


finalpatch:
This bazel issue has been open for almost 5 years.  I think it's very unlikely we can count on them resolving it.

paulo-coutinho:
Scala rules don't support it.

A hack to make it work on discussion in google groups:

```
I didn't realize this was for the scala rules. I'm not too sure how the scala rules work, but it doesn't look like they support adding things to the manifest directly. I was going to suggest using the "resources" attribute and add a MANIFEST.MF file there, but it doesn't look like singlejar (the tool that's used to create deploy jars) will merge MANIFEST.MF files. You probably want to file a feature with the scala rules: https://github.com/bazelbuild/rules_scala/issues Or perhaps there's some way to do it that they can point out.

As a workaround, you could add this manually in a genrule, it's pretty hacky though:

genrule(
  name = "scala_main_with_implementation_version",
  srcs = [":scala_main_deploy.jar"],
  outs = ["scala_main_with_implementation_version_deploy.jar"],
  executable = True,
  cmd = """
cp "$<" "$@"
chmod +w "$@"
unzip "$@" META-INF/MANIFEST.MF
# singeljar appears to write an extra CR LF at the end, and java
# doesn't appear to like more entries after that,
# so remove the last CR LF
truncate -s -2 META-INF/MANIFEST.MF
echo -n "Implementation-Version: 1.2.3.4\r\n" >> META-INF/MANIFEST.MF
zip -u "$@" META-INF/MANIFEST.MF
""",
)

A lot of the hackiness is due to the "truncate -s -2", there are probably more robust ways to remove the extra CR LF, especially since java will accept CR LF, CR, or LF for newlines in the manifest.
```

finalpatch:
Relying on string manipulation doesn't seem very robust.  Since this is not a crucial feature, I wouldn't mind leave it out. Thanks for your enthusiasm though.

paulo-coutinho:
I made a solution for my project. I save a file called .version when i download the binary from djinni, and use the data inside this file, instead of execute --version from binary.

PR #51, opened 2022-01-13
Remove const support in wasm codegen

bkotsopoulossc:
Currently there is support in the wasm codegen for enums and consts to run some global initializers that calls out to Javascript to define some constant values on the top-level Module object. This allows accessing enum values and constants by name in the module object. 

However, these types are already defined in the Typescript codegen. They can be imported and used directly without touching the module object. They can even be inlined at compile-time if the runtime conversions (between string and int) are not needed.

It may be convenient to also define enums and consts on the module object, but it has downsides. It bloats the JS generated by emscripten, and this bloat can't be stripped at build time. It means more imports to process during compilation time. It also slows down initialization time to run all of the global initializers.

This PR is just a sketch, but I propose removing the overhead incurred by defining consts/enums in the module and rely on typescript. Plain JS use cases are free to pass in hard-coded values, but we shouldn't invent a type system to support plain JS, especially when it adds an overhead for typescript use cases.

### DIFF
--- a/src/source/WasmGenerator.scala
+++ b/src/source/WasmGenerator.scala
@@ -159,101 +159,10 @@ class WasmGenerator(spec: Spec) extends Generator(spec) {
     }
   }
 
-  private def generateWasmConstants(w: IndentWriter, ident: Ident, consts: Seq[Const]) {
-    val nsprefix = spec.cppNamespace.replaceAll("::", "_")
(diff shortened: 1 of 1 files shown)


finalpatch:
This is now implemented in https://github.com/Snapchat/djinni/pull/53

PR #53, opened 2022-01-14
Make wasm enums and constants optional in codegen

finalpatch:
Add a command line flag `--wasm-omit-constants` which suppresses the generation of enum and constant values in wasm. They are still accessible in TypeScript, so if you only use TypeScript, you can set this flag to `true`.

Tested the generated code still works with the TypeScript test suite.

### DIFF
--- a/src/source/WasmGenerator.scala
+++ b/src/source/WasmGenerator.scala
@@ -230,30 +230,34 @@ class WasmGenerator(spec: Spec) extends Generator(spec) {
     val helper = helperClass(ident)
     writeHppFileGeneric(spec.wasmOutFolder.get, helperNamespace(), wasmFilenameStyle)(ident.name, origin, refs.hpp, Nil, (w => {
       w.w(s"struct $helper: ::djinni::WasmEnum<$cls>").bracedSemi{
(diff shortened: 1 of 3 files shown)


src/source/WasmGenerator.scala
```diff
@@ -230,30 +230,34 @@ class WasmGenerator(spec: Spec) extends Generator(spec) {
     val helper = helperClass(ident)
     writeHppFileGeneric(spec.wasmOutFolder.get, helperNamespace(), wasmFilenameStyle)(ident.name, origin, refs.hpp, Nil, (w => {
       w.w(s"struct $helper: ::djinni::WasmEnum<$cls>").bracedSemi{
-        w.wl("static void staticInitialize();");
+        if (!spec.wasmOmitConstants) {
```
bkotsopoulossc, line 233: I can't comment on the exact line, but no need for the mutex include on L228 without static init

bkotsopoulossc (approved):

src/source/WasmGenerator.scala
finalpatch: The refs are put in cpp files.  They are no longer generated so it doesn't matter.

PR #55, opened 2022-01-19
Add static method language annotation

finalpatch:
Static method declaration can now have `+c +j +o +w` annotations at the end (right before the comma).

If no annotation is present, then Djinni will generate binding for the static method in all languages.  eg. `static foo() +j;` will only generate java binding for `foo()`.

The absent of `+c` does not disable the static method in the generated C++ header file, but if a static method has only `+c` then all other language bindings are disabled.


bkotsopoulossc (approved):
Seems reasonable, would be great to have some tests

PR #57, opened 2022-01-20
Add test for static method language annotation

finalpatch:
If codegen is correct, then the java, objc and wasm tests should all build and pass without requiring the static method implementation in C++.


src/source/TsGenerator.scala
```diff
@@ -253,7 +253,7 @@ class TsGenerator(spec: Spec) extends Generator(spec) {
         case r: Record => generateRecord(td.origin, td.ident, td.doc, td.params, r, w)
         case i: Interface => {
           generateInterface(td.origin, td.ident, td.doc, td.params, i, w)
-          if (i.methods.exists(_.static)) {
+          if (i.methods.exists(m => m.static && m.lang.js)) {
```
bkotsopoulossc, line 256: Great example of why tests matter. The earlier we catch the issues, the more time we save

test-suite/generated-src/inFileList.txt
```diff
@@ -20,6 +20,7 @@ djinni/varnames.djinni
 djinni/relative_paths.djinni
 djinni/array.djinni
 djinni/nscopying.djinni
+djinni/static_method_language.djinni
```
bkotsopoulossc, line 23: Any purpose of having the infileList and the outfileList? I'm guessing it was copied from the original project, should we just delete them and stop setting the flags in the shell script?

bkotsopoulossc (approved):

PR #58, opened 2022-01-24
Allow usage of all caps identifiers in Djinni files

ggilles:
Originally, djinni idl files are meant to be written entirely in `snake_case`, and the ident style of the target would define the final result.

In practice, `CamelCase` identifiers in djinni files are preserved, so people started using them. See [examples](https://github.com/Snapchat/djinni/blob/master/test-suite/djinni/data_ref_view.djinni) [in the](https://github.com/Snapchat/djinni/blob/master/test-suite/djinni/function_prologue.djinni) [test-suite](https://github.com/Snapchat/djinni/blob/master/test-suite/djinni/single_language_interfaces.djinni).

Our own djinni files use the same convention as we use for Java/C++: `FooBar` for types and files, `fooBar` for fields, and `ALL_CAPS` for constants.

```
ExampleInterface = interface {
  exampleMethod();
  const EXAMPLE_CONSTANT: i32 = 0;
}
```

This style is certainly very common and leads to the same identifiers being generated for C++, Java and Objective-C in the general case.

Thus it appeared surprising that in this example, the constant would get translated in Objective-C as:

```
int32_t const ExampleInterfaceEXAMPLECONSTANT= 0;
```
Which is difficult to read, and doesn't adhere to standard Objective-C naming guidelines.

After the change suggested in this PR, the `EXAMPLE_CONSTANT` above gets translated as:

```
int32_t const ExampleInterfaceExampleConstant = 0;
```

Which is consistent with standard objective-c guidelines.

The final result means that for this fairly common style, the C++, Java and Objective-C identifiers are identical, except for Objective-C constants which are adjusted to be consistent with standard Objective-C guidelines.

### DIFF
--- a/src/source/generator.scala
+++ b/src/source/generator.scala
@@ -111,7 +111,22 @@ package object generatorTools {
     if (s.isEmpty) s else ", " + s
   }
   def q(s: String) = '"' + s + '"'
-  def firstUpper(token: String) = if (token.isEmpty()) token else token.charAt(0).toUpper + token.substring(1)
+
+  def leadingUpper(token: String) = {
+    if (token.isEmpty()) {
+      token
+    } else {
+      val head = token.charAt(0)
+      val tail = token.substring(1)
+      // Preserve mixed case identifiers like 'XXFoo':
+      // Convert tail to lowercase only when it is full uppercase.
+      if (tail.toUpperCase == tail) {
+        head.toUpper + tail.toLowerCase
+      } else {
+        head.toUpper + tail
+      }
+    }
+  }
 
   type IdentConverter = String => String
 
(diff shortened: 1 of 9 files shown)


finalpatch (approved):

finalpatch:
Nice,  thanks.  I don't see any problem with this change given the codegen for snake_case symbols are not affected.

finalpatch:
Sorry I had to revert this because it causes some unnecessary change to the C++ codegen.  In our codebase we have symbols such as `SSL`. They are not turned into `Ssl` which is arguably worse than before.

If you could find a way to scope this change to only objc and hide it behind an indent style flag then I'll be happy to merge it.

ggilles:
@finalpatch Thanks a lot for that feedback, I create of another approach which will be opt-in.

PR #60, opened 2022-01-27
Strict ident convert

ggilles:
This is a follow-up to https://github.com/Snapchat/djinni/pull/58 and the rational didn't change.

The solution, however, did.

There, I introduce so-called strict ident styles, that is, normal ident style suffixed with an exclamation-mark.

Specifying ident-style "FooBar!" will apply the CamelCase ident-style strictly, i.e. it will convert `snake_case` identifiers as usual, but `ALL_CAPS` will be converted correctly too.


finalpatch:
Change looks good.  There is a build failure though.

finalpatch:
One question, what is the motivation of using full caps in djinni files instead of just all_snake_case and let the the ident style take care of each language?

ggilles:
The motivation is 2 folds:

* avoid surprising behavior: we write djinni files in CamelCase and everything happens as if no transformation is done, *except* for constants that we write in all caps.
* having the exact same identifiers across languages allows easier grep/find. If we use all_snake_case for constant, that will be the only place in our repo where the identifier is written like that.

I'll fix the test later.

finalpatch (approved):
Thanks for making this opt-in and including tests.

PR #61, opened 2022-01-28
fix typo in generated comment

paulo-coutinho:


finalpatch:
thank you

finalpatch:
I'm in a bit of dilemma on this one.  On one hand, you are absolutely right this is a typo.  On the other hand,  this will change every single generated file so users will see a massive diff.

paulo-coutinho:
Is better fix now that is in early public stage than after with a lot of people using, this is normal

paulo-coutinho:
We'll merge these changes soon. Having diff is normal and git is for that. If we have to fix something, we will fix it soon.

finalpatch:
@bholmes-sc  what do you think? Should we leave this typo (in comments)  to avoid massive diffs or fix it now?

paulo-coutinho:
> @bholmes-sc what do you think? Should we leave this typo (in comments) to avoid massive diffs or fix it now?

the diff is only is a comment, the users don't have to do nothing

bholmes-sc (approved):

PR #62, opened 2022-01-28
add support for version

paulo-coutinho:
Hi,

With this you can check the version with command:

```
bazel run //src:djinni -- --version
```

or:

```
bazel run //generator:generator -- --version
```

If you are using the generated binary direct:

```
./djinni --version
```

It is a resource from scopt library.
It let you check what version you are using.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -24,20 +24,6 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-${{ env.cache-name }}-master
 
-      - name: Get the version
-        id: get_version
-        run: |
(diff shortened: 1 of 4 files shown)


paulo-coutinho:
I can think in a way to send the target version to be replaced there. What you think?

finalpatch:
I do not like this solution because of how it modifies source files in a github action.  May I ask why it is so important to embed the version string in the generator tool?  IMO it's enough to verify the hash of the downloaded package to ensure the right version is used.

paulo-coutinho:
I made a solution for my project. I save a file called .version when i download the binary from djinni, and use the data inside this file, instead of execute --version from binary.

PR #64, opened 2022-01-29
fix java enum

paulo-coutinho:
Hi,

This fix a bug in java generator:

```
domain/Customer.java:85: error: incompatible types: CustomerStatusEnum cannot be converted to int
        out.writeInt((int)this.mStatus);
                              ^
```


finalpatch (approved):

finalpatch:
Thanks! We have never used the Parcel feature so this bug has been unnoticed.

PR #66, opened 2022-02-01
default values for lib include

paulo-coutinho:
Hi,

This PR let users don't need add `<gen>BaseLibIncludePrefix` using a default value instead, that is the most common usage.

Thanks.


paulo-coutinho:
Don't make sense, since the paths are different for everyone

Issue #67, opened 2022-02-02
Wasm/TS: Nullability is not enforced across the interface


bkotsopoulossc:
When passing interfaces as function parameters, the generated typescript assumes every interface is *not nullable*, but the generated C++ code passes `std::sharedPtr<T>` types for each interface, which are *inherently nullable*. The typescript then has no way to legally check if the parameter is null or not (or undefined).

Let's say we have a djinni interface like this:
```
interface Foo +c {
  // ...
}

interface Bar +w {
  doSomething(foo: Foo);
}
```

On the C++ side, you call the web function and pass in an implementation of the C++ interface:
```
// .cpp
barObject->doSomething(std::make_shared<FooImpl>(...));
```

But it's also perfectly legal to pass in a nullptr:
```
// .cpp
barObject->doSomething(nullptr);
```

On the Typescript side, you may have some code like this to be defensive and check for the null pointer at runtime:
```
// .ts
const bar: Bar = {
  doSomething: (foo: Foo) => {
    if (!foo) { // this line fails type checking
      throw new Error()
    }
  }
}
```

But the if check fails type checking / linting because as far as typescript is concerned, that parameter can not be null/undefined, and the if check is always true.

We should either admit that interface types passed from c++ -> JS can be null, and codify that into the typescript signatures by generating them like `doSomething: (foo: (Foo | null)) => void`. Or, come up with some way to prevent null `shared_ptr` from crossing the boundary. This could be a runtime check, or it could be using a different type type that prevents construction from null ([example](https://github.com/microsoft/GSL/blob/main/include/gsl/pointers#L84)).

CC @LiFengSC

finalpatch:
You are right. Interfaces should be nullable and this is how it works in Java and ObjC too. Unless you specified `--cpp-nn-type` to prevent `nullptr` from entering the C++ pointer.

bholmes-sc:
My proposal here would be to first implement a runtime check for this anytime we make a call from C++ to TS.  
Don't we have some sort of proxy on the TS side that can verify if params are non-null?

paulocoutinhox:
It was merged and solved. Can be closed, no? Thanks.

PR #68, opened 2022-02-02
Fix TS interface nullability

finalpatch:
* Interface nullability is always retained in typescript regardless of whether cpp-nn-type is passed to command line. This is different from Java and ObjC (interface nullability is only retained when cpp-nn-type is provided)
* When cpp-nn-type is not available, a runtime null check is generated in the wasm marshaller classes's `fromCpp()` method
* When cpp-nn-type is not available, generate a `/*not-null*/` comment for non-optional interfaces in C++, as a reminder.
* null interface is represented with `undefined` in TS to be consistent with optional records.
* Demos and tests are fixed to respect nullability.
* TS projects are fixed to compile in strict mode

In the long term we want to be able to retain interface nullability in Java and ObjC too, but this will be a breaking change. We have to first audit all our existing methods that passes interfaces and make sure nullable interfaces are marked by optional<>. 

This address https://github.com/Snapchat/djinni/issues/67

### DIFF
--- a/examples/generated-src/cpp/sort_items.hpp
+++ b/examples/generated-src/cpp/sort_items.hpp
@@ -18,7 +18,7 @@ public:
     /** For the iOS / Android demo */
     virtual void sort(sort_order order, const ItemList & items) = 0;
 
-    static std::shared_ptr<SortItems> create_with_listener(const std::shared_ptr<TextboxListener> & listener);
+    static /*not-null*/ std::shared_ptr<SortItems> create_with_listener(const /*not-null*/ std::shared_ptr<TextboxListener> & listener);
 
     /** For the localhost / command-line demo */
     static ItemList run_sort(const ItemList & items);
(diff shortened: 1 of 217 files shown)


bkotsopoulossc (approved):
The fix makes sense. It's definitely going to be a big breaking change and make a lot of code more verbose. Have we thought about having a way to pass non-null interfaces as parameters? 9 times out of 10 we don't want nullability, but now we'll have a ton of checks for null.

CC @benjaminhollissc

finalpatch:
It's possible to make Djinni generate nonnull interfaces by default but we'll have to provide a C++ pointer type that can prevent `nullptr` getting in.

If we do that then nullable interfaces can be represented in djinni as `optional<I>` and they will be translated to plain `shared_ptr<>` in C++.

This approach will give us consistent behavior between records and interfaces (nonnull by default, nullable if optional<>). However,  changing all the C++ implementation classes to use a nonnull pointer is a non-trivial task.

bholmes-sc (approved):

Issue #71, opened 2022-02-22
WASM: Handling exceptions in JS


bkotsopoulossc:
**Background**

Emscripten as a framework supports [two ways](https://emscripten.org/docs/porting/exceptions.html) to build / run code that uses c++ exceptions. 

One is called "emscripten exception-handling" / emscripten-eh. In this method, C++ exceptions thrown actually bridge out to JS, and use JS to unwind the stack. The thing that ends up getting thrown is a JS number, which corresponds to the memory address of the exception_ptr on the wasm heap. The emscripten-generated JS code proxies each c++ method call and has special logic for handling catching errors of type `number`.

The other is called "wasm exception-handling" / wasm-eh. This is a new wasm proposal, where the runtime gets support to do its own stack unwinding, without relying on JS. In this implementation, you can throw and catch directly in wasm without bridging out to JS. When a C++ exception bubbles out of wasm and back to JS, the exception type is still a number, and it still corresponds to the memory address of the exception on the wasm heap. This new proposal is only supported in [recent browser versions](https://webassembly.org/roadmap/).

Since both of these implementations send a pointer back out to JS, the [recommendation](https://emscripten.org/docs/porting/Debugging.html#handling-c-exceptions-from-javascript) from emscripten to handle these errors is to expose a wasm function where you can take the `number`, cast it back to a pointer, access the exception object, and return the message back out to JS as a string. 

This works, but it means you need special logic in every JS catch block that is reachable from a c++ exception bubbling out. it's difficult to make sure this logic exists everywhere it is needed, and without it, debugging a caught exception is impossible, as the raw number has no useful information as it is. And, web apps sometimes rely on global uncaught error handlers to log and report errors that bubble all the way up to the root level (this could be the window, or it could be the root of a call stack from something like `setTimeout`).

**Change**

Djinni already has a proxy layer for JS calls to interfaces implemented in C++. It doesn't have a proxy layer for static methods, but this could be added.

This proxy layer could be used as a spot to catch C++ exceptions before they bubble out to JS, and transform them into canonical JS errors. 

[...]

PR #72, opened 2022-02-28
Translate C++ exceptions to JS Error objects

finalpatch:
Currently C++ exceptions are thrown into JS space as numbers.  JS code has to call a function to retrieve the error messages.  This is inconvenient and out of place. 

There are a few other behaviors that makes exceptions not as natural as it could be:

* JS Error passing through C++ code turns into C++ exception and when handled by upper level JS code, the original Error object is lost.
* Same happens to C++ exceptions passing through JS code and handled by upper level C++ code

This PR 
* changes the wasm codegen to catch C++ exceptions thrown in C++ implementation methods, and convert them to JS Error objects.
* Store the original JS Error object with the C++ exception so that it can be unwrapped when thrown back into JS land.
* Store the C++ exception pointer with the JS Error object so that it can be unwrapped when thrown back into C++

We also append a "C++: " prefix to the error message to make it easier to differentiate c++ exceptions in JS code.

This addresses https://github.com/Snapchat/djinni/issues/71


src/source/WasmGenerator.scala
```diff
@@ -365,13 +365,22 @@ class WasmGenerator(spec: Spec) extends Generator(spec) {
...
+            w.w("try").braced {
+              if (!m.ret.isEmpty) w.w("auto r = ")
+              if (m.static) w.w(s"$cls::") else w.w("self->")
+              writeAlignedCall(w, s"""${idCpp.method(m.ident)}(""", m.params, ")", p => {
+                s"${helperClass(p.ty.resolved)}::toCpp(${stubParamName(p.ident)})"
+              })
+              w.wl(";")
+              m.ret.fold()(r => w.wl(s"return ${helperClass(r.resolved)}::fromCpp(${cppMarshal.maybeMove("r", r)});"))
+            }
+            w.w("catch(const std::exception& e)").braced {
```
bkotsopoulossc, line 377: This seems reasonable for exceptions that were natively from C++. But for exceptions that originated in JS, can we re-create the original JS error (as best as we can)? We have this [JS exception](https://github.com/Snapchat/djinni/blob/11e3bc97ab59819da1254c67ce6fab1f3fa2bd27/support-lib/wasm/djinni_wasm.cpp#L76), which we can store more information in beyond just the message. We could read the relevant fields out of the JS one, hold them in the C++ one, then put them back in the JS one here when we re-throw

finalpatch: good idea.  I think I'm going to try to store the JS Error object inside the C++ exception, so that later I can unwrap it and let JS code retrieve the original error object.

bkotsopoulossc: That would be neat. Let's try it and see if it works. Preserving the correct stack trace as best as possible would be awesome. So that we're not losing information as the error bubbles up

support-lib/wasm/djinni_wasm.cpp
```diff
@@ -201,10 +187,18 @@ EM_JS(void, djinni_register_name_in_ns, (const char* prefixedName, const char* n
         ns[name] = Module[prefixedName];
 });
 
+void djinni_throw_native_exception(const std::exception& e) {
+    static auto throwCppException = em::val::module_property("djinni_throw_cpp_exception");
+    if (const auto* jsEx = dynamic_cast<const JsException*>(&e)) {
+        throwCppException(jsEx->cause());
```
bkotsopoulossc, line 193: The val type has a throw() method, I believe you can do:
```suggestion
        jsEx->cause().throw();
```

support-lib/wasm/djinni_wasm.cpp
```diff
@@ -180,6 +162,10 @@ EM_JS(void, djinni_init_wasm, (), {
             Module.protobuf[name] = proto;
         };
 
+        Module.djinni_throw_cpp_exception = function(e) {
+            throw (e instanceof Error) ? e : new Error("djinni: " + e);
```
bkotsopoulossc, line 166: With the below suggestion I believe you can do:
```suggestion
            throw new Error("djinni: " + e);
```

I also think that using the terms "native" or "wasm" might be easier to recognize for folks who aren't familiar with djinni.

test-suite/handwritten-src/js/CppExceptionTest.js
```diff
@@ -14,9 +20,18 @@ class CppExceptionTest {
...
+
+    testJsExceptionPassthrough() {
+        var thrown = null;
+        try {
+            this.cppInterface.throwJsException(new JsThrower());
+        } catch (e) {
+            thrown = e;
+        }
+        assertTrue(thrown instanceof Error);
+        assertEq(thrown.message, "JS Error");
```
bkotsopoulossc, line 34: We've got tests here of catching a C++ error in JS and a JS error in JS (passing through C++).

What do you think about some tests for catching a JS error in C++ and a C++ error in C++ (passing through JS)?

bkotsopoulossc (commented):
This approach looks good

support-lib/wasm/djinni_wasm.cpp
LiFengSC: Thanks, this looks much nicer

test-suite/handwritten-src/js/CppExceptionTest.js
LiFengSC: Yeah.  Also tests should check the C++ stack unwinding is performed.

LiFengSC: Actually,  I think we could probably do the same pass-through trick for Java code, too. In a follow-up PR.

LiFengSC:
@bkotsopoulossc I have made the requested changes:
* Use `val::throw()`
* Replace exception tag "djinni:" with "C++:"
* Support C++ exceptions passing through JS code
* More tests verifying all throw/catch combinations (c++->js, js->c++, c++->c++, js->js)

support-lib/wasm/djinni_wasm.cpp
```diff
@@ -188,11 +190,15 @@ EM_JS(void, djinni_register_name_in_ns, (const char* prefixedName, const char* n
...
-    static auto throwCppException = em::val::module_property("djinni_throw_cpp_exception");
     if (const auto* jsEx = dynamic_cast<const JsException*>(&e)) {
-        throwCppException(jsEx->cause());
+        jsEx->cause().throw_();
     } else {
-        throwCppException(em::val(e.what()));
+        static std::exception_ptr exptr;
+        static auto ErrorClass = em::val::global("Error");
+        auto error = ErrorClass.new_(std::string("C++: ") + e.what());
+        exptr = std::current_exception();
```
bkotsopoulossc, line 199: Technically this is the same exception as `e` but this gives us the ptr? We could also use `std::make_exception_ptr(e)` right? Looks good, just making sure I follow

support-lib/wasm/djinni_wasm.cpp
```diff
@@ -60,7 +60,13 @@ const em::val& JsProxyBase::_jsRef() const {
 
 void JsProxyBase::checkError(const em::val& v) {
     if (v.instanceof(em::val::global("Error"))) {
-        throw JsException(v);
+        auto cppExceptionPtr = v["cppExceptionPtr"];
```
bkotsopoulossc: Small nit, but any of these fields we introduce internal to djinni, maybe we prefix them with _djinni_ or something? Might make it clearer for other code that touches these objects

support-lib/wasm/djinni_wasm.cpp
```diff
@@ -60,7 +60,13 @@ const em::val& JsProxyBase::_jsRef() const {
...
+        if (!cppExceptionPtr.isUndefined()) {
+            std::exception_ptr* exptr = reinterpret_cast<std::exception_ptr*>(cppExceptionPtr.as<int>());
+            std::rethrow_exception(*exptr);
+        } else {
+            throw JsException(v);
+        }
```
bkotsopoulossc, line 69: Yes! This is exactly why I was prompting to add the tests for the extra cases, it didn't seem like we had handled this case

bkotsopoulossc (approved):

bkotsopoulossc (approved):

support-lib/wasm/djinni_wasm.cpp
LiFengSC: I already changed this to "_djinni_cpp_exception_ptr"

support-lib/wasm/djinni_wasm.cpp
LiFengSC: If I have an exception object that hasn't been thrown yet, I can use `make_exception_ptr` to get an `exception_ptr`.
If the exception is already thrown(like in our case), then `current_exception()` will return the `exception_ptr`.

support-lib/wasm/djinni_wasm.cpp
bkotsopoulossc: nice, missed the last commit

support-lib/wasm/djinni_wasm.cpp
```diff
@@ -201,10 +189,22 @@ EM_JS(void, djinni_register_name_in_ns, (const char* prefixedName, const char* n
         ns[name] = Module[prefixedName];
 });
 
+void djinni_throw_native_exception(const std::exception& e) {
+    if (const auto* jsEx = dynamic_cast<const JsException*>(&e)) {
```
bkotsopoulossc, line 193: @LiFengSC By the way, is this our first usage of RTTI in djinni? If it is, is this worth taking the dependency on it? We don't need RTTI for exceptions.

Could we generate code that catches both types, without using rtti?

```
    } catch(const JsException& e) {
      // re-throw JS
    } catch(const std::exception& e) {
        // re-throw native
    }
```

bkotsopoulossc: Actually I guess matching the catch clause requires RTTI anyways? Just wondering here really

LiFengSC: There are other uses of RTTI, eg. here we use it to detect whether the interface is a JS object or C++ object.
```
    static em::val _toJs(const std::shared_ptr<I>& c) {
        if (c == nullptr) {
            // null object
            return em::val::undefined();
        }
        else if (auto* p = dynamic_cast<JsProxyBase*>(c.get())) {
            // unwrap existing js proxy
            return p->_jsRef();
        } else {
            return GetOrCreateCppProxy<Self, void>()(c);
        }
    }
```

bkotsopoulossc: We should develop a better understanding of the size and perf overhead we add by relying on RTTI CC @bholmes-sc

PR #79, opened 2022-04-21
Add option to mark Obj-C symbols with "default" visibility

kersson:
We generate dynamic frameworks from djinni-generated Obj-C code, but we compile our code with `-fvisibility=hidden`, so we need to mark djinni-generated records and constants with "default" visibility.

This PR adds a `--objc-default-visibility-annotation` option to mark records and constants with "default" visibility:

```objc
__attribute__((visibility("default")))
```

This ensures that they will have external linkage even when compiled with `-fvisibility=hidden`.

### DIFF
--- a/src/source/ObjcGenerator.scala
+++ b/src/source/ObjcGenerator.scala
@@ -30,6 +30,8 @@ import scala.collection.parallel.immutable
 
 class ObjcGenerator(spec: Spec) extends BaseObjcGenerator(spec) {
 
+  val visibilityAnnotation = if (spec.objcDefaultVisibilityAnnotation) "__attribute__((visibility(\"default\"))) " else ""
+
   class ObjcRefs() {
     var body = mutable.TreeSet[String]()
     var header = mutable.TreeSet[String]()
@@ -109,7 +111,7 @@ class ObjcGenerator(spec: Spec) extends BaseObjcGenerator(spec) {
       if(i.methods.exists(_.static)) {
         val protocolHeader = "#import " + q(spec.objcIncludePrefix + marshal.headerName(ident))
         writeObjcFile(marshal.implHeaderName(ident), origin, List(protocolHeader), w => {
-          w.wl(s"@interface $self : NSObject<$self>")
+          w.wl(s"${visibilityAnnotation}@interface $self : NSObject<$self>")
           for (m <- i.methods) {
             if (m.static && m.lang.objc) {
               w.wl
@@ -128,14 +130,14 @@ class ObjcGenerator(spec: Spec) extends BaseObjcGenerator(spec) {
     writeObjcFile(marshal.headerName(ident), origin, refs.header, w => {
       for (c <- i.consts if marshal.canBeConstVariable(c)) {
         writeDoc(w, c.doc)
-        w.w(s"extern ")
+        w.w(s"${visibilityAnnotation}extern ")
         writeObjcConstVariableDecl(w, c, self)
         w.wl(s";")
       }
 
       w.wl
       writeDoc(w, doc)
-      if (useProtocol(i.ext, spec)) w.wl(s"@protocol $self <NSObject>") else w.wl(s"@interface $self : NSObject")
+      if (useProtocol(i.ext, spec)) w.wl(s"@protocol $self <NSObject>") else w.wl(s"${visibilityAnnotation}@interface $self : NSObject")
 
       for (m <- i.methods) {
         if (!m.static || (!spec.objcGenProtocol && m.lang.objc)) {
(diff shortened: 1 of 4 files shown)


kersson:
I wanted to add tests for this new option, but I didn't see a place to readily do so. Is there support for testing code generated with various non-standard options?

src/source/ObjcGenerator.scala
```diff
@@ -128,14 +130,14 @@ class ObjcGenerator(spec: Spec) extends BaseObjcGenerator(spec) {
     writeObjcFile(marshal.headerName(ident), origin, refs.header, w => {
       for (c <- i.consts if marshal.canBeConstVariable(c)) {
         writeDoc(w, c.doc)
-        w.w(s"extern ")
+        w.w(s"${visibilityAnnotation}extern ")
```
LiFengSC, line 133: I wonder if constants visibility should be declared in `BaseObjcGenerator.writeObjcConstMethImpl`?

src/source/ObjcGenerator.scala
```diff
@@ -249,7 +251,7 @@ class ObjcGenerator(spec: Spec) extends BaseObjcGenerator(spec) {
         w.wl
         for (c <- r.consts if marshal.canBeConstVariable(c)) {
           writeDoc(w, c.doc)
-          w.w(s"extern ")
+          w.w(s"${visibilityAnnotation}extern ")
           writeObjcConstVariableDecl(w, c, noBaseSelf);
```
LiFengSC, line 255: And this too.

LiFengSC (commented):
Thank you for your contribution.  You are right we currently lack a means to test codegen by text comparison. Although given how straightforward this one is, I'm okay to leave the test for now until we have something better to support this kind of tests.

One question I have is about the annotation before constants.  You currently have them at the forward declaration in the header files, but I wonder if it should be in the implementation file, where the constants are actually instantiated.

src/source/ObjcGenerator.scala
kersson: My understanding is that visibility attributes should be placed in declarations not definitions. Here's a quote from the GCC wiki (https://gcc.gnu.org/wiki/Visibility):
>In your header files, wherever you want an interface or API made public outside the current DSO, place `__attribute__ ((visibility ("default")))` in struct, class and function declarations you wish to make public (it's easier if you define a macro as this). You don't need to specify it in the definition.

kersson:
>You are right we currently lack a means to test codegen by text comparison.

Even stronger than text comparison, ideally I'd love to write a test for this particular option by actually compiling the generated code into a dynamic shared library with `-fvisibility=hidden` and then verifying that the symbols remain visible!

src/source/ObjcGenerator.scala
LiFengSC: I see.  I was thinking for constants,  adding to `writeObjcConstMethImpl` will need only 1 change than the declarations where 2 changes are needed.

LiFengSC:
@kersson would it also work if you build the djinni generated code as a separate lib with `visibility=default` and the rest of the code with `visibility=hidden`.  That way we probably don't even need this command line option?

src/source/ObjcGenerator.scala
LiFengSC: maybe, but writeObjcConstVariableDecl has 3 callers,  2 with `extern`, 1 without.

kersson:
> @kersson would it also work if you build the djinni generated code as a separate lib with `visibility=default` and the rest of the code with `visibility=hidden`. That way we probably don't even need this command line option?

Interesting idea! My main concern is that it adds complexity to our project, which is already a massive code base. We build djinni-generated code with a couple different build systems, so we'd have to add in many separate modules/rules to build djinni-generated code separately, with different visibility settings. For all our other public APIs, we already explicitly mark our interfaces as exported/publicly visible in the headers, so we'd be making a special case for djinni-generated code.

But I understand that the tradeoff for this project is needing to maintain an additional option going forward.

I'll play around with this to see if this approach could work for us.

src/source/ObjcGenerator.scala
kersson: Right, so I'd have to add an argument to indicate whether the declaration is in the header (possibly add visibility) or the implementation (don't add visibility)? Not sure if it's worth it in this case, but I don't have strong feelings either way.

LiFengSC:
> > @kersson would it also work if you build the djinni generated code as a separate lib with `visibility=default` and the rest of the code with `visibility=hidden`. That way we probably don't even need this command line option?
> 
> Interesting idea! My main concern is that it adds complexity to our project, which is already a massive code base. We build djinni-generated code with a couple different build systems, so we'd have to add in many separate modules/rules to build djinni-generated code separately, with different visibility settings. For all our other public APIs, we already explicitly mark our interfaces as exported/publicly visible in the headers, so we'd be making a special case for djinni-generated code.
> 
> But I understand that the tradeoff for this project is needing to maintain an additional option going forward.
> 
> I'll play around with this to see if this approach could work for us.

Yeah I would prefer not to introduce new options if this can be achieved relatively easily without it.  AFAIK we do this (build djinni generated files as a seprate lib) at Snap.  If you use a capable build system (eg. cmake, bazel) building part of the source as a library is fairly easy.

kersson:
We do use bazel, but due to the way we bundle our generated sources alongside everything else we'd either need to make somewhat messy exclusion rules or reorganize things. None of which is difficult, but given that this option works well for us without any other changes to our source tree structure or build rules, we'll probably just use our own fork with this option added.

Thanks for the discussion!

PR #81, opened 2022-04-28
Add whenAll() and whenAny() function for C++ futures

LiFengSC:
Add a few utility functions to make handling a group of futures together easier in C++

`whenAll()` returns a new future that becomes ready when all of the supplied futures are ready.
`whenAny()` returns a new future that becomes ready when the first future in the supplied futures becomes ready.


sanjayrd (commented):
haven't reviewed code yet, but wanted to note that we should replace <redacted> with this

bholmes-sc (approved):

PR #87, opened 2022-06-07
Fix warnings in support-lib/java

remyjette:
* Future.java:83: warning: [overloads] then(FutureHandler<T>) in Future
  is potentially ambiguous with <R>then(FutureHandlerWithReturn<T,R>) in
  Future

  Since both of these are interfaces, a class could implement both and
  then the overload of `then()` to call would be ambiguous. In practice
  this would not happen as these objects are created via lambdas and the
  functional interface, so suppressed the warning.

* Future.java:71: warning: [rawtypes] found raw type: AtomicReference

  Fixed these by changing `new AtomicReference(...)` to `new
  AtomicReference<>(...)`

* NativeFutureHandler.java:19: warning: [rawtypes] found raw type:
  FutureHandler

  Added type arguments

* Outcome.java:39: warning: [rawtypes] found raw type: Outcome

  Added wildcard type arguments

* Outcome.java:46: warning: [rawtypes] found raw type: Class

  Added wildcard type argument

### DIFF
--- a/support-lib/java/com/snapchat/djinni/Future.java
+++ b/support-lib/java/com/snapchat/djinni/Future.java
@@ -57,10 +57,12 @@ public class Future<T> implements java.util.concurrent.Future<T> {
     }
     
     // Handler routine for type U that does not return a value
+    @FunctionalInterface
     public interface FutureHandler<U> {
         public void handleResult(Future<U> res) throws Throwable;
     }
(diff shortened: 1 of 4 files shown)


LiFengSC (approved):

LiFengSC:
Thanks Remy!  Changes look good.

PR #92, opened 2022-07-01
revert <NSObject> in generated protocols

LiFengSC:
Revert the <NSObjet> requirement on Djinni generated ObjC protocols, based on the discussion in [this thread](https://github.com/cross-language-cpp/djinni-generator/issues/133) and consulting with iOS developers.

Basically, the lack of <NSObject> does not prevent us from achieving anything in ObjC (it just makes the user code slightly more verbose in the worst case). However, it does make the generated code considerably more difficult to consume in Swift code.

### DIFF
--- a/examples/generated-src/objc/TXSTextboxListener.h
+++ b/examples/generated-src/objc/TXSTextboxListener.h
@@ -5,7 +5,7 @@
 #import <Foundation/Foundation.h>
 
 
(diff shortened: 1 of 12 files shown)


benjaminhollissc (approved):

paulocoutinhox:
Hi,

With NSObject my code is breaking when inherit.

image

PR #94, opened 2022-07-20
Support optional<> in ObjC containers

LiFengSC:
ObjC containers (NSArray, NSSet, NSDictionary) do not allow `nil` as elements.  However, the codegen does not prevent this, so if C++ puts empty optionals in the container, it will raise as a runtime error in ObjC.

This PR detects empty optional<> in C++ containers, and converts them into `NSNull` objects before returning to ObjC. And converts `NSNull` objects back to empty optionals when they are passed from ObjC to C++.

The check is only performed in containers with optional elements, so the performance of normal containers with non-null elements are unaffected.

### DIFF
--- a/support-lib/objc/DJIMarshal+Private.h
+++ b/support-lib/objc/DJIMarshal+Private.h
@@ -233,7 +233,7 @@ public:
     using Boxed = Optional;
 
     static CppType toCpp(ObjcType obj) {
-        if (obj/* && (id)obj != [NSNull null]*/) {
+        if (obj) {
             return T::Boxed::toCpp(obj);
         } else {
             return CppType();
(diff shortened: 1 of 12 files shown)


paulocoutinhox:
Nice. Thanks.

LiFengSC:
@bholmes-sc what do you think?

paulocoutinhox:
When you merge it, can you create a new version?

bholmes-sc (approved):
Change looks good.   Are we able to check this change for any performance impact?

LiFengSC:
> Change looks good. Are we able to check this change for any performance impact?

There is no performance impact when the type is not `optional<>` because the null checked code path is not used.

When the type is `optional<>`, it will null check and this comes with some overhead, but otherwise it won't work correctly anyway (will crash previously if a null is passed).  So performance wise I don't think it regresses in any way.

PR #98, opened 2022-08-11
[Future] Suppose resolve/reject convenience methods

bkotsopoulossc:
It would be nice to have some convenience methods for dealing with futures. 

We often see code like this:
```
::djinni::Future<int> SomeClass::someFunc() {
  // this
  if (!someCondition) {
    ::djinni::Promise<int> promise;
    auto future = promise.getFuture();
    promise.setValue(-1);
    return future;
  }

  // or this
  if (!someCondition) {
    ::djinni::Promise<int> promise;
    auto future = promise.getFuture();
    promise.setException(std::runtime_error("failed"));
    return future;
  }
  
  // or this
  try {
    auto data = getData();
    // ...
  } catch (...) {
    ::djinni::Promise<int> promise;
    auto future = promise.getFuture();
    promise.setException(std::current_exception());
    return future;
  }
}
```

It's a bit verbose, and we can borrow some inspiration from Javascript Promises.

This PR introduces some static convenience methods for returning a future with a value or error.

### DIFF
--- a/support-lib/cpp/Future.hpp
+++ b/support-lib/cpp/Future.hpp
@@ -108,6 +108,49 @@ class PromiseBase {
 public:
     Future<T> getFuture();
 
+    // Use to immediately resolve a promise and return the resulting future.
+    // Useful for returning from trivial cases in functions that return djinni::Future<>
+    // Examples:
+    //   return Promise<int>::resolve(4);
+    //   return Promise<std::string>::resolve(std::move(someStr));
+    static Future<T> resolve(const typename ValueHolder<T>::type& val) {
+        PromiseBase<T> promise;
+        promise.setValue(val);
+        return promise.getFuture();
+    }
+    static Future<T> resolve(typename ValueHolder<T>::type&& val) {
+        PromiseBase<T> promise;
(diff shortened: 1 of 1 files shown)


LiFengSC:
How about have these methods in the `Future<T>` type instead of `Promise<T>`?

This is because you end up returning an object of `Future<T>`, so IMO it makes sense to have the factory method in `Future<T>`.

LiFengSC:
Otherwise the change looks good and makes sense.

bkotsopoulossc:
We can, we just need to manually create the shared state to create the future. We can hide it all in the static method on the class

LiFengSC:
> We can, we just need to manually create the shared state to create the future. We can hide it all in the static method on the class

Yeah I know we can't just create a future by itself.  I mean we just move these same static methods to the Future<> class because I think it feels more natural from the user's perspective.

LiFengSC (approved):
Actually,  scrap what I said earlier.  I can see where you come from,  in Javascript the methods are called Promise.resolve() and Promise.reject() so it makes sense to do the same in Promise.

PR #99, opened 2022-08-12
update bazel scala rules (for java 18)

LiFengSC:
On M1 Macs, only java 18 can successfully complete compilation and testing of unit tests.  However, the current bazel scala rules do not work in java 18.

This updates scala rules to work in java 18.

Also tested in java 8 (compiles scala and runs the tool, but won't compile unit tests on M1 macs).

### DIFF
--- a/bzl/scala_config.bzl
+++ b/bzl/scala_config.bzl
@@ -1,4 +1,4 @@
 load("@io_bazel_rules_scala//:scala_config.bzl", "scala_config")
 
 def djinni_scala_config():
-    scala_config()
+    scala_config(scala_version = "2.11.12")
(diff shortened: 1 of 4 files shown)


bholmes-sc (approved):

bzl/scala_config.bzl
```diff
@@ -0,0 +1,4 @@
+load("@io_bazel_rules_scala//:scala_config.bzl", "scala_config")
+
+def djinni_scala_config():
+    scala_config(scala_version = "2.11.12")
```
bkotsopoulossc, line 4: Why do we have a new bazel file instead of putting this in setup_deps?

finalpatch:
It does not work Otherwise. Took me a while to figure out using a new file
solves the dependency issue.

Issue #101, opened 2022-08-23
[Future] [Wasm] Catch exceptions and reject promise


bkotsopoulossc:
## Motivation

Suppose I have an async function that returns a djinni future, and for whatever reason an exception is thrown on the synchronous path (before kicking off any async code paths).

```
djinni::future<int> foo() {
  if (someThing) {
    throw std::runtime_error("failed");
  }

  djinni::Promise<int> promise;
  auto future = promise.get_future();

  // ... promise is used on an async call path and set to a value or exception

  return future;
}
```

This throw site could be changed to:
```
if (someThing) {
  return djinni::Promise::reject(std::runtime_error("failed"));
}
```

But suppose we were calling some synchronous code that we don't control, and it throws inside. We could try-catch around it, but it's still something the programmer has to remember:
```
try {
  checkSomeThing();
} catch (const std::exception& e) {
  return djinni::Promise::reject(std::current_exception());
}
```

## Ask
Specifically for JS < -> Wasm, can we have a wrapper that catches exceptions thrown when invoking an async function, and return them as a rejected promise instead? We already have wrappers that catch exceptions to convert between C++ and JS errors, but they aren't aware of async functions with futures available.

## What does JS do?

If we have the following JS code ([link](https://www.typescriptlang.org/play?ssl=24&ssc=1&pln=1&pc=1#code/IYZwngdgxgBAZgV2gFwJYHsL3egFAShgG8AoGcmZACwCd0B3GCAU0YFEa6bcAiOYVABtmAEx74A3CQC+JEohQYsNJAHVU1dAmQBBegOQFiZCshphjFK9jyST5aTCjBkUKjFzNCpaxSiYQdGEAOkF0AHNeFQh1TW09AwAuJ2AEcKpkHgAaGC8pK1lZElBIWAUoNEwYaNiqBI0jH1NzS18YYH0NGwJ8ikdnV3dPb3trfwhAkLDInhqNOs7kZOc0jOzcuwKZOTnqesM7El2qLV1FnpIgA) to playground), with an async function throwing, the exception is implicitly captured into the state of the promise, and only is re-thrown if the promise is awaited.
```
async function foo() {
    throw new Error("failed");
}

function runWithoutAwait() {
    try {
        foo();
    } catch (e) {
        console.log("runWithoutAwait: caught", e);
    }
}

async function runWithAwait() {
    try {
        await foo();
    } catch (e) {
        console.log("runWithAwait: caught", e);
    }
}

[...]

PR #102, opened 2022-08-24
Translate C++ exceptions to async Errors in async functions

LiFengSC:
Change the codegen for wasm function stubs from:
```
    try {
        auto r = call_into_wasm_implementation_method(...);
        return ::djinni::TranslatorClass<...>::fromCpp(std::move(r));
    }
    catch(const std::exception& e) {
        djinni::djinni_throw_native_exception(e);
        throw;
    }
```
to
```
    try {
        auto r = call_into_wasm_implementation_method(...);
        return ::djinni::TranslatorClass<...>::fromCpp(std::move(r));
    }
    catch(const std::exception& e) {
        return djinni::ExceptionHandlingTraits<::djinni::TranslatorClass<...>>::handleNativeException(e);
    }
```

Add new type `ExceptionHandlingTraits` in wasm support library.  The default implementation is calling `djinni_throw_native_exception(e)` like before.

Specialize `ExceptionHandlingTraits` in the Future translator to return the error as a rejected JS promise:
```
        auto r = FutureAdaptor<U>::NativePromiseType::reject(std::current_exception());
        return FutureAdaptor<U>::fromCpp(std::move(r));
```

This addresses https://github.com/Snapchat/djinni/issues/101

New test cases are added to verify this works as expected.

### DIFF
--- a/support-lib/cpp/Future.hpp
+++ b/support-lib/cpp/Future.hpp
@@ -179,6 +179,7 @@ protected:
     }
 private:
     detail::SharedStatePtr<T> _sharedState = std::make_shared<SharedState<T>>();
+    detail::SharedStatePtr<T> _sharedStateReadOnly = _sharedState; // allow calling getFuture() after setValue()
 
     template <typename UpdateFunc>
     void updateAndCallResultHandler(UpdateFunc&& updater) {
@@ -362,7 +363,7 @@ private:
 
 template <typename T>
 Future<T> detail::PromiseBase<T>::getFuture() {
-    return Future<T>(std::atomic_load(&_sharedState));
+    return Future<T>(_sharedStateReadOnly);
 }
 
 template <typename U>
--- a/support-lib/wasm/Future_wasm.hpp
+++ b/support-lib/wasm/Future_wasm.hpp
@@ -105,4 +105,12 @@ public:
     }
 };
 
+template<typename U>
+struct ExceptionHandlingTraits<FutureAdaptor<U>> {
+    static em::val handleNativeException(const std::exception& e) {
+        auto r = FutureAdaptor<U>::NativePromiseType::reject(std::current_exception());
+        return FutureAdaptor<U>::fromCpp(std::move(r));
+    }
+};
+
 } // namespace djinni
--- a/support-lib/wasm/djinni_wasm.hpp
+++ b/support-lib/wasm/djinni_wasm.hpp
@@ -602,6 +602,30 @@ private:
 extern "C" void djinni_register_name_in_ns(const char* prefixedName, const char* namespacedName);
 extern "C" void djinni_throw_native_exception(const std::exception& e);
 
+template<typename U>
+struct DefaultInit {
+    static U get() { return U{}; }
+};
+template<>
+struct DefaultInit<em::val> {
+    static em::val get() { return em::val::undefined(); }
+};
+
+template<typename T>
+struct ExceptionHandlingTraits {
+    using R = typename T::JsType;
+    static R handleNativeException(const std::exception& e) {
+        djinni_throw_native_exception(e);
+        return DefaultInit<R>::get();
+    }
+};
+template<>
+struct ExceptionHandlingTraits<void> {
+    static void handleNativeException(const std::exception& e) {
+        djinni_throw_native_exception(e);
+    }
+};
+
 template<typename ClassType>
 class DjinniClass_ : public em::class_<ClassType> {
 public:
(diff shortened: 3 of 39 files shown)


test-suite/handwritten-src/ts/AsyncTest.ts
```diff
@@ -65,6 +65,17 @@ class AsyncTest extends TestCase {
         const s = await this.m.testsuite.TestHelpers.checkAsyncComposition(new AsyncInterfaceImpl());
         assertEq(s, "42");
     }
+
+    async testEarlyThrow() {
+        const r = this.m.testsuite.TestHelpers.asyncEarlyThrow();
+        let s = null;
+        try {
+            await r;
```
LiFengSC, line 73: @bkotsopoulossc 
Exception is thrown here in `await` rather than in the async call above.

bholmes-sc (approved):

support-lib/cpp/Future.hpp
```diff
@@ -179,6 +179,7 @@ class PromiseBase {
     }
 private:
     detail::SharedStatePtr<T> _sharedState = std::make_shared<SharedState<T>>();
+    detail::SharedStatePtr<T> _sharedStateReadOnly = _sharedState; // allow calling getFuture() after setValue()
```
bkotsopoulossc, line 182: Perfect

test-suite/handwritten-src/ts/AsyncTest.ts
bkotsopoulossc: Amazing

support-lib/wasm/Future_wasm.hpp
```diff
@@ -105,4 +105,12 @@ class FutureAdaptor
     }
 };
 
+template<typename U>
+struct ExceptionHandlingTraits<FutureAdaptor<U>> {
+    static em::val handleNativeException(const std::exception& e) {
+        auto r = FutureAdaptor<U>::NativePromiseType::reject(std::current_exception());
```
bkotsopoulossc, line 111: How come we use current_exception here instead of e? Should we do that in the other specializations of ExceptionHandlingTraits too?

support-lib/wasm/djinni_wasm.hpp
```diff
@@ -602,6 +602,30 @@ class GenericBuffer: public DataObject {
...
+struct DefaultInit<em::val> {
+    static em::val get() { return em::val::undefined(); }
+};
+
+template<typename T>
+struct ExceptionHandlingTraits {
+    using R = typename T::JsType;
+    static R handleNativeException(const std::exception& e) {
+        djinni_throw_native_exception(e);
+        return DefaultInit<R>::get();
```
bkotsopoulossc, line 619: Hmm I'm a bit confused about this line. For non-async functions, we used to have code like:
```
} catch (const std::exception& e) {
        djinni::djinni_throw_native_exception(e); 
        throw;
}
```

I think we still want that re-throw for non-native exceptions for non-async functions right?

support-lib/wasm/Future_wasm.hpp
LiFengSC: Because when we call this function:
```
    template <typename E>
    void setException(const E& ex) {
        setException(std::make_exception_ptr(ex));
    }
```

E will be deduced to `std::exception` .

`make_exception_ptr` has this signature:

```
template< class E >
std::exception_ptr make_exception_ptr( E e ) noexcept;
```
And it will slice our object into a `std::exception`, losing all the context data we put in it.

LiFengSC: Basically `setException` should not be called with a polymorphic reference.  It needs to be the actual type to avoid exception being sliced.

support-lib/wasm/djinni_wasm.hpp
LiFengSC: `djinni::djinni_throw_native_exception(e); ` throws a JS exception and will not return. 
The line after that was just to satisfy the compiler.

bkotsopoulossc: Ahh got it, thanks https://cs.github.com/Snapchat/djinni/blob/5accefb308e9b958b67b2e533e5b0dc2d30eef4e/src/source/WasmGenerator.scala?q=djinni_throw_native_exception#L378-L381

support-lib/wasm/Future_wasm.hpp
bkotsopoulossc: Sounds good, I guess just keeping it consistent helps but I understand the constraint now

bkotsopoulossc (approved):

PR #111, opened 2022-10-24
Simplify nested namespaces codegen

ukreator:
Using `namespace ns1::ns2::ns3 {` instead of `namespace ns1 { namespace ns2 { namespace ns3 {` which slightly simplifies codegen and makes the code conformant with [modernize-concat-nested-namespaces](https://clang.llvm.org/extra/clang-tidy/checks/modernize/concat-nested-namespaces.html) clang-tidy check.

### DIFF
--- a/src/source/generator.scala
+++ b/src/source/generator.scala
@@ -361,11 +361,10 @@ abstract class Generator(spec: Spec)
     ns match {
       case "" => f(w)
       case s =>
-        val parts = s.split("::")
-        w.wl(parts.map("namespace "+_+" {").mkString(" ")).wl
+        w.wl(s"namespace $s {").wl
         f(w)
         w.wl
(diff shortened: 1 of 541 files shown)


src/source/generator.scala
```diff
@@ -361,11 +361,10 @@ abstract class Generator(spec: Spec)
     ns match {
       case "" => f(w)
       case s =>
-        val parts = s.split("::")
-        w.wl(parts.map("namespace "+_+" {").mkString(" ")).wl
+        w.wl(s"namespace $s {").wl
```
ukreator, line 364: the only change

LiFengSC: Thanks!  I agree it's reasonable to generate C++17 at this point.

LiFengSC (approved):

Issue #112, opened 2022-10-29
32 bit flag to enum generation issue


martin00001313:
In case of defining flags which later are translated to 32 bit unsigned enum, for the 32th enum value issue is occurred as '1<<31' is considered signed i32 operation and is overflow.
Instead, by generating with unsigned postfix 'u', i.e. '1u<<31' will work fine.

Proposal, use '1u<<shift' instead of '1 << shift'. Both C++ and ObjC support it fine

similar: https://github.com/cross-language-cpp/djinni-generator/issues/142

paulocoutinhox:
Closed by https://github.com/Snapchat/djinni/pull/114

PR #113, opened 2022-10-29
Use of all 32 bits for flags -> enum mapping of C++/ObjC - #112

martin00001313:
Allow use of all 32 bits for definition of enum flags as '1<<31' leads to overflow as it considered i32 shift.
Instead '1u<<31' should be used which will not consider the upper bit as signed bit

### DIFF
--- a/test-suite/generated-src/cpp/access_flags.hpp
+++ b/test-suite/generated-src/cpp/access_flags.hpp
@@ -18,7 +18,7 @@ enum class access_flags : unsigned {
     SYSTEM_READ = 1u << 6,
     SYSTEM_WRITE = 1u << 7,
     SYSTEM_EXECUTE = 1u << 8,
-    EVERYBODY = (1u << 9) - 1,
+    EVERYBODY = 0 | OWNER_READ | OWNER_WRITE | OWNER_EXECUTE | GROUP_READ | GROUP_WRITE | GROUP_EXECUTE | SYSTEM_READ | SYSTEM_WRITE | SYSTEM_EXECUTE,
 };
 constexpr access_flags operator|(access_flags lhs, access_flags rhs) noexcept {
     return static_cast<access_flags>(static_cast<unsigned>(lhs) | static_cast<unsigned>(rhs));
--- a/test-suite/generated-src/cpp/access_flags.hpp
+++ b/test-suite/generated-src/cpp/access_flags.hpp
@@ -9,16 +9,16 @@ namespace testsuite {
 
 enum class access_flags : unsigned {
     NOBODY = 0,
-    OWNER_READ = 1 << 0,
-    OWNER_WRITE = 1 << 1,
-    OWNER_EXECUTE = 1 << 2,
-    GROUP_READ = 1 << 3,
-    GROUP_WRITE = 1 << 4,
-    GROUP_EXECUTE = 1 << 5,
-    SYSTEM_READ = 1 << 6,
-    SYSTEM_WRITE = 1 << 7,
-    SYSTEM_EXECUTE = 1 << 8,
-    EVERYBODY = (1 << 9) - 1,
+    OWNER_READ = 1u << 0,
+    OWNER_WRITE = 1u << 1,
+    OWNER_EXECUTE = 1u << 2,
+    GROUP_READ = 1u << 3,
+    GROUP_WRITE = 1u << 4,
+    GROUP_EXECUTE = 1u << 5,
+    SYSTEM_READ = 1u << 6,
+    SYSTEM_WRITE = 1u << 7,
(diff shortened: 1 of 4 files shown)


test-suite/generated-src/cpp/access_flags.hpp
```diff
@@ -9,16 +9,16 @@ namespace testsuite {
...
+    OWNER_READ = 1u << 0,
+    OWNER_WRITE = 1u << 1,
+    OWNER_EXECUTE = 1u << 2,
+    GROUP_READ = 1u << 3,
+    GROUP_WRITE = 1u << 4,
+    GROUP_EXECUTE = 1u << 5,
+    SYSTEM_READ = 1u << 6,
+    SYSTEM_WRITE = 1u << 7,
+    SYSTEM_EXECUTE = 1u << 8,
+    EVERYBODY = (1u << 9) - 1,
```
LiFengSC, line 21: I think we'd need to fix this as well to make the last bit truly usable.

LiFengSC: I mean the EVERYBODY case should be `|` of all members rather than `(1u << N) - 1`, which will overflow if N is 32.

martin00001313: yes, I was expected it to be the same as is done for [cross-platform-djinni](https://github.com/cross-language-cpp/djinni-generator/releases/tag/v1.3.2)
Btw, just updated the logic of 'all' flag based on bitwise operation instead of using subtraction

LiFengSC: Awesome! Thank you!

LiFengSC: Could you check in all the regenerated files including objc/ts/wasm files?

LiFengSC: There's a CI check that verifies  all generated files to be up to date.

martin00001313: I've just updated the generated enum files(for which the tests were failing)
there is a problem related to ts, as for ts there is no 'u' support
Seems we need to have a different enum generation for ts w/o 'u'

LiFengSC: I wonder if we can simply change the base type of the enum to int32_t.  I tested that this passes the compilation and the bit flags are correct.

LiFengSC: and since it uses signed value, it's available across all platforms that djinni needs to support (including java and js where unsigned is n/a).

martin00001313: by changing it to int32 you reserve the first bit for sign, which will decrease the capacity if enum: i.e. 31 instead of 32

LiFengSC: We do need the all flags fix, otherwise 1<<32 will not compile

martin00001313: the tricky point is related to ts, maybe we can have a separate logic of enum generation for it?

LiFengSC: no the first bit is interpreted as sign only when you view it as an integer,  but if you use it as bit flags (checking/comparing with bit operations) then it does not matter.

LiFengSC:
```
#include <iostream>
#include <bitset>

enum class MyFlags1: int32_t {
    NO_FLAGS = 0,
    FLAG1 = 1 << 30,
    FLAG2 = 1 << 31,
    ALL_FLAGS = 0 | FLAG1 | FLAG2,
};

int main (int argc, char *argv[])
{
    auto a = MyFlags1::FLAG2;
    std::cout << std::bitset<32>(static_cast<int32_t>(a)) << std::endl;
    std::cout << (a == MyFlags1::FLAG2) << std::endl;
}
```
output: 
```
10000000000000000000000000000000
1
```

LiFengSC: and ALL_FLAGS is 11000000000000000000000000000000
which is also correct.

LiFengSC: It fails the compilation in the original code when the base type is unsigned only because when the sign bit is 1, the number value becomes negative, which cannot be casted to unsigned.

martin00001313: yes the reason is casting. just conceptually use of signed bit is a annoying and if I was the owner of codebase, I'd avoid to have it.

btw, i32_t from functional point of view is also ok. So, as a 'good to go' - use int32_t and update 'ALL' prop's calculation(by using bitwise or instead of -)

PR #114, opened 2022-11-08
Allow use of all 32 bits as enum flags for ObjC and C++

martin00001313:
- Allow use of all 32 bits as enum flags for ObjC and C++
- Re-design 'ALL' flag's value calculation base on bitwise operation instead of subtraction
- As enum flag type is used int32_t (for C++) to emphasize memory usage and do not have a platform dependency(unlike int)
- As enum flag type for ObjC is used signed integer: NSInteger

### DIFF
--- a/src/source/CppGenerator.scala
+++ b/src/source/CppGenerator.scala
@@ -65,7 +65,7 @@ class CppGenerator(spec: Spec) extends Generator(spec) {
       refs.hpp.add("#include <functional>") // needed for std::hash
     }
 
-    val flagsType = "unsigned"
+    val flagsType = "int32_t"
     val enumType = "int"
     val underlyingType = if(e.flags) flagsType else enumType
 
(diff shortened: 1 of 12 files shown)


martin00001313:
On local laptop I had tmp issue related to bazel so why skipped test flow execution locally. Btw, published the latest version with correction of test suites

LiFengSC:
> On local laptop I had tmp issue related to bazel so why skipped test flow execution locally. Btw, published the latest version with correction of test suites

Thanks, I'll review this later today.

LiFengSC:
There still seems to be an 'unsigned' in CppMarshal.scala

LiFengSC:
CppMarshal.scala
96:50:            val underlyingType = if(e.flags) " : unsigned" else ""

LiFengSC:
You shouldn't need to manually change the generated code

LiFengSC:
I'll do some test on this branch and then we can merge this.  Thanks @martin00001313 !

LiFengSC (approved):

PR #116, opened 2022-11-11
Allow functions that return Futures to become coroutines

LiFengSC:
This improves the existing coroutine support to make it fully bidirectional.  Now it can turn any function that returns a djinni::Future<> into a coroutine.  I modified a few test functions to exercise this new feature.


bkotsopoulossc (approved):
Great thanks, much more useful

PR #119, opened 2022-11-15
Use java8 CompletableFuture as djinni future

LiFengSC:
Use CompletableFuture instead of our own Promise/Future implementation.

This should allow better integration with the Java/Kotlin ecosystem. 

This requires Android API level 24.


LiFengSC:
Close because of API 24 requirement.

Issue #75, opened 2022-03-16
future<void>


LiFengSC:
Currently `future<>` type requires a non-void type, because `void` is not a valid Djinni type. This means the IDL cannot represent a future with no value, but only a completion status.

bkotsopoulossc:
+1, this is quite awkward without

LiFengSC:
* The C++ `Future` can already support `Future<void>`
* Java has `java.lang.Void` that can be used to box a void value.
* What is ObjC's equivalent of `java.lang.Void`?
* Javascript  `Promise` also already supports `Promise<void>`.

bkotsopoulossc:
Seems like everything is easy except obj-c. I don't think they have a boxed version of void. Do you need a boxed type? Maybe NSValue or a void* or something? What are the constraints?

paulocoutinhox:
Hi,

In obj-c you have type `id`:
https://riptutorial.com/objective-c/example/16005/id

I don't know if it help.

The `java.lang.Void` can be integer?
https://stackoverflow.com/questions/10839042/what-is-the-difference-between-java-lang-void-and-void

Thanks.

nolanw:
For Objective-C, maybe `NSNull`?

nmtitov:
`NSNull` is the only option for objc I guess since we can't put `void` in box

bkotsopoulossc:
to get the full value of https://github.com/Snapchat/djinni/pull/116 we really need to get a void type, otherwise it's painful to interop coroutines that return void with djinni futures that can't

PR #120, opened 2022-11-18
support future<void>

LiFengSC:
This addresses
https://github.com/Snapchat/djinni/issues/75

Type mapping of future<void>:
* C++,  `Future<void>`
* Java, `Future<Void>`
* ObjC, `DJFuture<NSNull *>`
* Javascript/Typescript, `Promise<void>`

The void type can only be used as the type parameter of futures.  The void type's `toCpp` and `fromCpp` methods are not implemented so trying to pass void objects as actual data will no work (compilation will fail).  There is a special case in the parser to allow writing `:void` as the return type, and this will be ignored.

### DIFF
--- a/support-lib/jni/Future_jni.hpp
+++ b/support-lib/jni/Future_jni.hpp
@@ -20,8 +20,6 @@
 #include "Marshal.hpp"
 #include "../cpp/Future.hpp"
 
-#include <iostream>
-
 namespace djinni {
 
 struct PromiseJniInfo {
@@ -105,7 +103,6 @@ public:
                 reinterpret_cast<NativePromiseType*>(nativePromise)
             };
             if (jex == nullptr) {
-                // promise->setValue(RESULT::Boxed::toCpp(jniEnv, static_cast<typename RESULT::Boxed::JniType>(jres)));
                 SetResult<RESULT>::setCppResult(jniEnv, *promise, static_cast<typename RESULT::Boxed::JniType>(jres));
             } else {
                 const auto& throwableJniInfo = JniClass<ThrowableJniInfo>::get();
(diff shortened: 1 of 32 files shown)


support-lib/jni/Future_jni.hpp
```diff
@@ -20,6 +20,8 @@
 #include "Marshal.hpp"
 #include "../cpp/Future.hpp"
 
+#include <iostream>
```
bkotsopoulossc, line 23: NIT - remove

support-lib/jni/Future_jni.hpp
```diff
@@ -83,7 +105,8 @@ class FutureAdaptor
                 reinterpret_cast<NativePromiseType*>(nativePromise)
             };
             if (jex == nullptr) {
-                promise->setValue(RESULT::Boxed::toCpp(jniEnv, static_cast<typename RESULT::Boxed::JniType>(jres)));
+                // promise->setValue(RESULT::Boxed::toCpp(jniEnv, static_cast<typename RESULT::Boxed::JniType>(jres)));
```
bkotsopoulossc, line 108: Still need these comments? And below on L144

bkotsopoulossc (approved):
Thank you

test-suite/handwritten-src/cpp/test_helpers.cpp
```diff
@@ -256,6 +256,10 @@ djinni::Future<std::string> TestHelpers::check_async_composition(const std::shar
 #endif
 }
 
+::djinni::Future<void> TestHelpers::void_async_method(djinni::Future<void> f) {
+    return f;
```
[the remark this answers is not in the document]
LiFengSC: yes

PR #122, opened 2022-12-04
Fix missing :: prefix for djinni namespace to avoid possible name lookup issues

ukreator:
A few calls to the supporting lib were missing global :: prefix in the WASM generator.

### DIFF
--- a/examples/generated-src/wasm/NativeSortItems.cpp
+++ b/examples/generated-src/wasm/NativeSortItems.cpp
@@ -21,7 +21,7 @@ void NativeSortItems::sort(const CppType& self, int32_t w_order,const em::val& w
              ::djinni_generated::NativeItemList::toCpp(w_items));
     }
     catch(const std::exception& e) {
-        return djinni::ExceptionHandlingTraits<void>::handleNativeException(e);
+        return ::djinni::ExceptionHandlingTraits<void>::handleNativeException(e);
     }
 }
 em::val NativeSortItems::create_with_listener(const em::val& w_listener) {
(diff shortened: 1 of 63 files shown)


paulocoutinhox:
Hi. Nice. What is the difference in practice?

ukreator:
I was looking into issues with name collisions when you define structs with the same name in different .djinni files with different `--wasm-namespace`. The generated classes for wasm are all placed into `namespace djinni_generated` and have `NativeStructName` for the struct name which means they cause issues for linker.

I hacked the wasm scala generator a bit to use different namespaces for these `Native*` structs and without prefixing the support lib function calls with `::` it causes compiler errors. So fixing it would be great for consistency and for potential name lookup issues when fixing the name collision issues (which I will defer to @LiFengSC )

LiFengSC:
> I was looking into issues with name collisions when you define structs with the same name in different .djinni files with different `--wasm-namespace`. The generated classes for wasm are all placed into `namespace djinni_generated` and have `NativeStructName` for the struct name which means they cause issues for linker.
> 
> I hacked the wasm scala generator a bit to use different namespaces for these `Native*` structs and without prefixing the support lib function calls with `::` it causes compiler errors. So fixing it would be great for consistency and for potential name lookup issues when fixing the name collision issues (which I will defer to @LiFengSC )

This is a separate issue from the one addressed in this PR, right?

LiFengSC (approved):

ukreator:
> This is a separate issue from the one addressed in this PR, right?

yeah, the fix is more involved

LiFengSC:
> > This is a separate issue from the one addressed in this PR, right?
> 
> yeah, the fix is more involved

Isn't that a problem in java and objc too?  All the marshaller classes are generated in the `djinni_generated` namespace with no module namespace names added.  Why haven't we had this problem in the past?

ukreator:
> Isn't that a problem in java and objc too? All the marshaller classes are generated in the djinni_generated namespace with no module namespace names added. Why haven't we had this problem in the past?

JNI/Obj-C++ ones seem to be fine - their namespaces are controlled by `--jni-namespace` and `--objcpp-namespace`. But the WASM generator uses the JNI namespace: https://github.com/Snapchat/djinni/blob/main/src/source/WasmGenerator.scala#L37

So if you don't specify `--jni-namespace` (e.g. you don't generate Android bindings) then you'll get name conflicts.

PR #124, opened 2022-12-08
Fix Promise optional in typescript

bkotzz:
We currently have a bug where we don't support futures with optionals for Typescript codegen. In particular, 
`foo: future<optional<int>>` generates a Typescript signature of `Promise<number>`.

This may have been omitted at the beginning with external types in general. However it works fine to have optional types in ES6 containers like Array/Set/Map ([example](https://www.typescriptlang.org/play?#code/MYewdgzgLgBAtgTwMoFNYF4ZhQdxqqAHmgCcBLMAcxgB8YBXMAExQDMKUmA+ACgG0ARExCUBAGgbM2HJgF0AlAChQkWIgCCJEgEMEMTNjyadCYlHJVaklu2zceA1iBDjr0uxIEAjbSQHyAbmVwaHgEAFltAAcARn0sXBhIqLMLajpGGxkJUgp0t1tOXj5FGDKYQScXTx8-WTFS8r5M904a3wF6xQUglVDEZIAmeMMk6NS8qxbCphzzSYypGeLGssrnV2mZLvKKrY8C7YaexWDIEAAbFAA6C5EeRAJAs4hLm7vKB4RjXWe+t9u9wG0RifxCAI+XyGzyAA)).

In this PR we update the codegen to take into account the nullability in the TS generator, then add a new test across the languages. CC @LiFengSC

### DIFF
--- a/src/source/TsGenerator.scala
+++ b/src/source/TsGenerator.scala
@@ -72,7 +72,7 @@ class TsGenerator(spec: Spec) extends Generator(spec) {
   }
 
   def toTsType(tm: MExpr, addNullability: Boolean = true): String = {
-    def args(tm: MExpr) = if (tm.args.isEmpty) "" else tm.args.map(f).mkString("<", ", ", ">")
+    def args(tm: MExpr) = if (tm.args.isEmpty) "" else tm.args.map(arg => toTsType(arg, addNullability)).mkString("<", ", ", ">")
     def f(tm: MExpr): String = {
       tm.base match {
         case MOptional =>
--- a/test-suite/generated-src/ts/test.ts
+++ b/test-suite/generated-src/ts/test.ts
@@ -451,6 +451,11 @@ export interface TestHelpers_statics {
     futureRoundtrip(f: Promise<number>): Promise<string>;
     asyncEarlyThrow(): Promise<number>;
     voidAsyncMethod(f: Promise<void>): Promise<void>;
+    /**
+     * If the input is empty, returns back an empty future.
+     * If the input is non-empty, returns back the value plus one.
+     */
+    addOneIfPresent(f: Promise<number | undefined>): Promise<number | undefined>;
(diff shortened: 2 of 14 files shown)


src/source/TsGenerator.scala
```diff
@@ -72,7 +72,7 @@ class TsGenerator(spec: Spec) extends Generator(spec) {
   }
 
   def toTsType(tm: MExpr, addNullability: Boolean = true): String = {
-    def args(tm: MExpr) = if (tm.args.isEmpty) "" else tm.args.map(f).mkString("<", ", ", ">")
+    def args(tm: MExpr) = if (tm.args.isEmpty) "" else tm.args.map(arg => toTsType(arg, addNullability)).mkString("<", ", ", ">")
```
bkotzz, line 75: This is the fix. The only difference between `f` and `toTsType` is it takes into account the optional logic
```
(if (addNullability) nullityAnnotation(tm) else "")
```

We capture `addNullability` from the outer scope, as it should be the same for all the args.

test-suite/generated-src/ts/test.ts
```diff
@@ -451,14 +451,19 @@ export interface TestHelpers_statics {
...
-    getOptionalSet(): Set<string>;
-    checkOptionalSet(os: Set<string>): boolean;
-    getOptionalMap(): Map<string, string>;
-    checkOptionalMap(om: Map<string, string>): boolean;
+    getOptionalList(): Array<string | undefined>;
+    checkOptionalList(ol: Array<string | undefined>): boolean;
+    getOptionalSet(): Set<string | undefined>;
+    checkOptionalSet(os: Set<string | undefined>): boolean;
+    getOptionalMap(): Map<string | undefined, string | undefined>;
+    checkOptionalMap(om: Map<string | undefined, string | undefined>): boolean;
```
bkotzz, line 466: This is an expected change to existing codegen

LiFengSC (approved):
Awesome! Thank you!

PR #129, opened 2023-01-22
Fix CI target branch

ggilles:
I noticed this problem while having a look at how tests work: continuous integration tests don't run when committing on the main branch.


LiFengSC (approved):
Thank you!

PR #133, opened 2023-02-02
[Wasm] Clean up promise rejection code

bkotsopoulossc:
We've seen some issues that manifest from rejecting a JS promise with a null value, or with a DOMException that does not have a stack property. In both cases we made some assumptions that aren't always true, so let's guard against them.

PTAL @LiFengSC

### DIFF
--- a/support-lib/cpp/Future.hpp
+++ b/support-lib/cpp/Future.hpp
@@ -174,6 +174,9 @@ protected:
     }
     template <typename E>
     void setException(const E& ex) {
+        // Let's try to prevent throwing non-exception types.
+        static_assert(std::is_convertible_v<E, std::exception>, "setException() called with type that is not convertible to std::exception");
+
         setException(std::make_exception_ptr(ex));
     }
     void setException(std::exception_ptr ex) {
--- a/support-lib/cpp/Future.hpp
+++ b/support-lib/cpp/Future.hpp
@@ -172,7 +172,6 @@ protected:
             sharedState->value = std::move(val);
         });
     }
-
     template <typename E>
     void setException(const E& ex) {
         setException(std::make_exception_ptr(ex));
--- a/support-lib/cpp/Future.hpp
+++ b/support-lib/cpp/Future.hpp
@@ -173,9 +173,8 @@ protected:
         });
     }
 
-    // Lock down to std::exception as a base class, as otherwise make_exception_ptr will happily take other
-    // pointers, and we don't want to encourage rejecting a promise with something other than an exception object.
-    void setException(const std::exception& ex) {
+    template <typename E>
+    void setException(const E& ex) {
         setException(std::make_exception_ptr(ex));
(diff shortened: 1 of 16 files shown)


support-lib/wasm/Future_wasm.hpp
```diff
@@ -34,28 +34,42 @@ class FutureAdaptor
...
+        if constexpr (std::is_void_v<CppResType>) {
+            pNativePromise->setValue();
         } else {
+            pNativePromise->setValue(RESULT::Boxed::toCpp(res));
+        }
+
+        delete pNativePromise;
+    }
+
+    static void rejectNativePromise(int context, em::val err) {
```
bkotsopoulossc, line 49: Splitting up the rejection path. We overcomplicated this by having two params in one. It creates a lot more edge cases than dealing with the single parameter that the promise was resolved or rejected with.

support-lib/cpp/Future.hpp
```diff
@@ -172,8 +172,10 @@ class PromiseBase {
             sharedState->value = std::move(val);
         });
     }
-    template <typename E>
-    void setException(const E& ex) {
+
+    // Lock down to std::exception as a base class, as otherwise make_exception_ptr will happily take other
+    // pointers, and we don't want to encourage rejecting a promise with something other than an exception object.
+    void setException(const std::exception& ex) {
```
finalpatch, line 178: This will slice the exception object and lose information.  `make_exception_ptr` makes a copy, and since the parameter type is std::exception&, only the std::exception part will be kept.  The template here is just an attempt to retain the full exception.

bkotsopoulossc: Hmm okay. What annoyed me with this template is I was able to pass in any pointer data. Like a string literal. I want the compiler to stop me. Any ideas how we can prevent that and not lose the derived class information?

bkotsopoulossc:
Really need to make the wasm tests easier to run. I keep getting 
```                                                                               
                This is not the tsc command you are looking for                
                                                                               
To get access to the TypeScript compiler, tsc, from the command line either:

- Use npm install typescript to first add TypeScript to your project before using npx
- Use yarn to avoid accidentally running code from un-installed packages
```

There should be one command that runs tests and it should run in CI

LiFengSC (approved):
lgtm

support-lib/cpp/Future.hpp
LiFengSC: Sorry that was me, accidentally commented in my personal account. 

We can tighten the parameter requirement by rejecting types that are not `std::is_convertible<std::exception>` with a static_assert.

bkotsopoulossc: okay I will try that out

bkotsopoulossc:
@LiFengSC added some tests and a static_asset on exception type. PTAL and if it's okay with you please merge

PR #139, opened 2023-05-05
Fix JNI marshalling of zero-length protobuf objects from native->Java

kthornberrysc:
Protobuf objects that contain only empty or default values take up zero bytes on the wire. In this case the JNI fromCpp() marshaller will try to map a DirectByteBuffer onto a zero-length nullptr C++ buffer, which is forbidden (https://docs.oracle.com/en/java/javase/20/docs/specs/jni/functions.html#newdirectbytebuffer).

On my Android installation with -Xcheck:jni enabled, it will abort with the error "JNI ERROR: non-nullable address is NULL". In other cases the behaviour is undefined.

Fix: check for zero-length C++ buffer and in that case use an empty Java array-backed ByteBuffer to pass to the Java deserializer.

### DIFF
--- a/support-lib/jni/Marshal.hpp
+++ b/support-lib/jni/Marshal.hpp
@@ -564,6 +564,11 @@ namespace djinni
         }
     };
 
+    struct ByteBufferJniInfo {
+        const GlobalRef<jclass> clazz{ jniFindClass("java/nio/ByteBuffer") };
+        const jmethodID method_allocate{ jniGetStaticMethodID(clazz.get(), "allocate", "(I)Ljava/nio/ByteBuffer;") };
+    };
+
     // 
     template<typename CPP_PROTO, typename JAVA_PROTO, typename JAVA_SERIALIZER = GpbMessageLiteSerializer>
     class Protobuf {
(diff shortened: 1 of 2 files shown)


support-lib/jni/Marshal.hpp
```diff
@@ -607,12 +612,22 @@ namespace djinni
         {
             // Serialize to C++ vector
             std::vector<uint8_t> cppBuf(c.ByteSizeLong());
```
LiFengSC, line 614: The size is `c.ByteSizeLong()`, so once we know the size is 0, there is no need to construct the `vector<>` any more.

support-lib/jni/Marshal.hpp
```diff
@@ -607,12 +612,22 @@ namespace djinni
...
+                [[maybe_unused]]
+                bool success = c.SerializeToArray(cppBuf.data(), static_cast<int>(cppBuf.size()));
+                assert(success);
+
+                // Wrap C++ vector as java.nio.ByteBuffer (no ownership)
+                javaBuf = LocalRef<jobject>(jniEnv, jniEnv->NewDirectByteBuffer(cppBuf.data(), cppBuf.size()));
+            } else {
+                // Can't have a zero-length DirectByteBuffer so use an empty Java array-backed ByteBuffer
+                const auto& byteBuffer = JniClass<ByteBufferJniInfo>::get();
+                javaBuf = LocalRef<jobject>(jniEnv, jniEnv->CallStaticObjectMethod(byteBuffer.clazz.get(), byteBuffer.method_allocate, 0));
```
LiFengSC, line 627: All zero size buffers are the same, so maybe we can create once into a static variable, and reuse it.

LiFengSC:
Thanks for the fix!

support-lib/jni/Marshal.hpp
kthornberrysc: The vector lifetime needs to extend beyond the size check to the `deserializeJavaProto()` call

LiFengSC (approved):

PR #148, opened 2023-06-30
Force bazel to use d8 instead of dx for Android

LiFengSC:
bazel 5 by default uses dx which is deprecated since 2018 and no longer available in buildtools v31 and later.  This change upgrade bazel version to lates v5 release (5.4.1) and added command line flags to .bazelrc to force it to use d8 instead of dx. The hard coded buildtools version number is removed.


bholmes-sc (approved):
We should also plan to upgrade to Bazel 6 soon.

LiFengSC:
> 

IIRC the scala toolchain does not support bazel 6

paulocoutinhox:
https://github.com/bazelbuild/rules_scala#bazel-compatible-versions

LiFengSC:
Tested on Ubuntu with sdk33 & ndk21d.
* build perftest 
* test djinni-java-tests
All good.

PR #153, opened 2023-11-19
Fix TS compilation error on tsc 4.5.4

mutagene:
Fixes  for error when running `bazel run //test-suite:server-ts` -

```
../../djinni/vendor/third-party/proto/ts/test.ts:3:23 - error TS2497: This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.

3 import * as Long from "long";
                        ~~~~~~

../../djinni/vendor/third-party/proto/ts/test2.ts:3:23 - error TS2497: This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.

3 import * as Long from "long";
```

### DIFF
--- a/test-suite/djinni/vendor/third-party/proto/ts/test.ts
+++ b/test-suite/djinni/vendor/third-party/proto/ts/test.ts
@@ -1,6 +1,6 @@
 /* eslint-disable */
 import { util, configure, Writer, Reader } from "protobufjs/minimal";
-import * as Long from "long";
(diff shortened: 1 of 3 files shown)


LiFengSC:
Thank you!  I'll run some tests with Snap's codebase and get back to you!

LiFengSC (approved):

Issue #107, opened 2022-09-09
NDK not seems not compatible 


coxfilur:
ERROR: /Users/lichunyu/work/djinni-main/WORKSPACE:13:23: fetching android_ndk_repository rule //external:androidndk: java.io.IOException: Could not read RELEASE.TXT in Android NDK: /Users/lichunyu/Library/Android/sdk/ndk/RELEASE.TXT (No such file or directory) Unable to read the Android NDK at /Users/lichunyu/Library/Android/sdk/ndk, the path may be invalid. Is the path in android_ndk_repository() or ANDROID_NDK_HOME set correctly? If the path is correct, the contents in the Android NDK directory may have been modified.

My ndk version:(in /Users/lichunyu/Library/Android/sdk/ndk)
21.3.6528147	21.4.7075529

LiFengSC:
I just tested and it works for me.  My $ANDROID_NDK_HOME is "~/Library/Android/sdk/ndk/21.3.6528147"

LiFengSC:
Also ndk in this repo is only used for building the android example app.   You can use any android sdk/ndk version to build your own app.

cucbin:
did you use Bazel to build? It seem Bazel error. you can edit your WORKSPACE and locate the NDK path to android_ndk_repository() like:
android_ndk_repository(name = "androidndk", path = "/path/to/ndk", api_level = 21)

or, the value can be omitted if `ANDROID_NDK_HOME` environment variable is set.

floriansegginger:
If the NDK is only used to build the Android example app, why is it part of the build step for Djinni? It just adds an unnecessary dependency to users of the library who might just need to generate code with Djinni without Android support.

LiFengSC:
@floriansegginger you are absolutely right.  I think the reason this hasn't bothered us at Snap is because the main purpose of using Djinni is to support Android and iOS development so pretty much all developers already have NDK installed. But on the other hand, the Java support in Djinni is not really tied to Android so NDK does not have to be a hard dependency.  When I have some time, I'll try if I could remove this dependency.  In the mean time, if you want to open a PR to fix this, I'm happy to review and merge it.

PR #154, opened 2023-11-21
Allow code generation without android SDK/NDK

sonicdebris:
We have the same issue mentioned in https://github.com/Snapchat/djinni/issues/107 (specifically: https://github.com/Snapchat/djinni/issues/107#issuecomment-1502909975).

In short, some of our developers working with the generator are not android developers, and would rather not have to set-up the android SDK/NDK to use djinni, considering those are only needed to build the android examples.

I know nothing about Bazel, but still I decided to try and find a solution to this. I started investigating a way to conditionally trigger the android repo rules for the examples only, when I found this discussion:
https://github.com/bazelbuild/rules_android_ndk/issues/92

From there, I understand that the rule _should_ only be applied in case one tries to build a target that requires the android SDK, but apparently that is not the case yet. In the same discussion, I found a suggestion for a workaround, which amounts to only defining the `android_sdk_repository` and `android_sdk_repository` rules if the environment variables for the paths to the SDK/NDK are set:
https://github.com/envoyproxy/envoy-mobile/pull/2039

I implemented the same, with a trivial tweak, which is printing a warning in case the environment variables are not set, to counterbalance the fact that the build failure message for the `android-app` and `android_sdk_repository` is now less explicit.

It seems to work fine, hopefully it's not too hacky and can be useful upstream.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -9,9 +9,10 @@ djinni_setup_deps()
 
 # --- Everything below is only used for examples and tests
 
-# android_sdk_repository fails to find build_tools if we don't explicitly set a version.
-android_sdk_repository(name = "androidsdk")
-android_ndk_repository(name = "androidndk", api_level = 21)
+load("//bzl:android_configure.bzl", "android_configure")
+android_configure(name = "local_config_android")
+load("@local_config_android//:android_configure.bzl", "android_workspace")
+android_workspace()
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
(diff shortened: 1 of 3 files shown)


LiFengSC:
Thanks!  This is something I have wanted for a long time.  I'll need to do some tests with Snap's codebase and then get back to you.

LiFengSC:
@sheldonneuberger-sc we use a trimmed down version of the workspace file, so I believe this will not impact Snap code.

LiFengSC:
I tested this on my laptop and it works very well! Thank you @sonicdebris !

sheldonneuberger-sc (approved):
Nice, thanks for fixing that.

nmtitov:
This is very good idea, I like it. I mostly do iOS development and having Android dependency installed is really annoying for me.

jsgfkkfhsppplscPpp-cloud (commented):
snapchat

PR #160, opened 2024-01-16
Composer generator

LiFengSC:
Code generator for bridging C++ and Composer/TypsScript.

The TypeScript type definitions reuses the TS generator for WebAssembly because they are 99% the same.

The C++ marshalling code targets the new Composer support library.


support-lib/composer/Future_composer.hpp
```diff
@@ -0,0 +1,96 @@
...
+  * Unless required by applicable law or agreed to in writing, software
+  * distributed under the License is distributed on an "AS IS" BASIS,
+  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
+  */
+
+#pragma once
+
+#include "djinni_composer.hpp"
```
bholmes-sc, line 19: is order important here?  if not sort them.

support-lib/composer/Outcome_composer.hpp
```diff
@@ -0,0 +1,63 @@
...
+  * distributed under the License is distributed on an "AS IS" BASIS,
+  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  * See the License for the specific language governing permissions and
+  * limitations under the License.
+  */
+
+#pragma once
+
+#include "djinni_composer.hpp"
+#include "../cpp/expected.hpp"
```
bholmes-sc, line 20: can we avoid the relative paths?

bholmes-sc (approved):

support-lib/composer/Outcome_composer.hpp
LiFengSC: removed relative paths.  they are not needed.  "cpp" is already in include search path.

PR #161, opened 2024-02-16
Fix exception propagation to JS for async void functions 

arekmula:
Hey,

We encountered an issue where exceptions were not properly propagated for functions returning `future<void>`. It's because future was never resolved in the `doResolve` for void types and exception was never raised there so it could be caught and translated using `djinni_native_exception_to_js`

### DIFF
--- a/support-lib/wasm/Future_wasm.hpp
+++ b/support-lib/wasm/Future_wasm.hpp
@@ -96,6 +96,7 @@ public:
         void doResolve() {
             try {
                 if constexpr (std::is_void_v<typename RESULT::CppType>) {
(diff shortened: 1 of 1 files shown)


LiFengSC (commented):
Looks good. Thank you!

LiFengSC:
The CI error is not from your PR.  Should be fixed with https://github.com/Snapchat/djinni/pull/162 if you pull that from `main`

arekmula:
I rebased on top of https://github.com/Snapchat/djinni/pull/162

LiFengSC (approved):

PR #163, opened 2024-02-23
Workaround for known GCC bug regarding explicit specialization in non namespace scope

arekmula:
Hey,

This is a workaround for a [known bug in GCC](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85282). 

[Similar issue on StackOverflow](https://stackoverflow.com/questions/49707184/explicit-specialization-in-non-namespace-scope-does-not-compile-in-gcc)

### DIFF
--- a/support-lib/cpp/Future.hpp
+++ b/support-lib/cpp/Future.hpp
@@ -383,15 +383,17 @@ public:
             this->_promise.setValue(std::forward<V>(value));
         }
     };
-    template <>
-    struct PromiseType<void>: PromiseTypeBase {
-        void return_void() {
-            this->_promise.setValue();
-        }
-    };
     using promise_type = PromiseType<T>;
 #endif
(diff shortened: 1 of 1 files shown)


support-lib/cpp/Future.hpp
```diff
@@ -383,16 +383,18 @@ class Future {
...
     using promise_type = PromiseType<T>;
 #endif
 };
 
+#if defined(DJINNI_FUTURE_HAS_COROUTINE_SUPPORT)
+template<> template<> struct Future<void>::PromiseType<void>:PromiseTypeBase {
+    void return_void() {
+        this->_promise.setValue();
+    }
+}
```
LiFengSC, line 395: missing `;`

arekmula: Fixed

PR #168, opened 2024-05-08
Fix compiler warnings in Future implementation

jb-gcx:
- Unused parameters caused warnings in my project, so I removed them
- I think `Future.get()` is intended to return a non-nullable - unless I'm missing some reason why it shouldn't?

### DIFF
--- a/support-lib/cpp/Future.hpp
+++ b/support-lib/cpp/Future.hpp
@@ -23,6 +23,7 @@
 #include <condition_variable>
 #include <mutex>
 #include <cassert>
+#include <exception>
 
 #ifdef __cpp_coroutines
 #if __has_include(<coroutine>)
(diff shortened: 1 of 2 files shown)


LiFengSC:
Hi, thanks for the PR.  While the compiler warning fix is great, I'm not 100% sure about nonnull on the ObjC future.get method.  I can see the "setValue" method also takes a "nullable", so in theory we may indeed get a null from get.

jb-gcx:
So if I understand correctly, in the ObjC version `nil` is always a possible result, regardless of the value type I use? I might argue that this should change, but I'll make that argument in a separate PR. I've removed the questionable commit from this PR 👍

LiFengSC (approved):

PR #174, opened 2024-06-29
Make JNI string conversions behave like ObjC string conversions

jb-gcx:
When a C++ std::string contains invalid UTF-8, the result is an empty string when trying to pass it to ObjC. For JNI, it caused an exception. For consistency, I would like encoding errors to result in empty strings in JNI as well.

### DIFF
--- a/support-lib/jni/djinni_support.cpp
+++ b/support-lib/jni/djinni_support.cpp
@@ -439,7 +439,12 @@ using WcharConverter = std::wstring_convert<std::codecvt_utf16<wchar_t, 0x10ffff
 using Utf8Converter = std::wstring_convert<std::codecvt_utf8_utf16<char16_t, 0x10ffff, std::codecvt_mode::little_endian>, char16_t>;
 
 jstring jniStringFromWString(JNIEnv * env, const std::wstring & str) {
-    std::string u16 = WcharConverter{}.to_bytes(str);
+    std::string u16{};
+    try {
+        u16 = WcharConverter{}.to_bytes(str);
+    } catch (...) {
+        // Conversion error - empty string will be returned
+    }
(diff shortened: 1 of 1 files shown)


LiFengSC:
Hi @jb-gcx,

Thanks for the contribution!  This is an area I had hesitated for a long time.

On one hand,  I kind of want to make programming errors stand out more so that they could be corrected earlier.  Passing invalid characters as string is one of them,  binary objects should be passed using the `binary` type.  Changing the string to something else (such as empty string) may allow this kind of errors to slip past testing. 

On the other hand,  I understand it could be annoying when simply passing a string argument may require the caller to handle potential encoding errors.

I can't decided which is better,  convenience or strictness.

Also,  if you just want to return a fallback string when there is an encoding error, `std::wstring_convert`'s constructor lets you pass `byte_error` and `wide_error` that will be used in case of error instead of throwing.

jb-gcx:
Thanks for the response @LiFengSC !

Regarding your worries about errors slipping past, I absolutely agree. I much prefer them to be visible in general.

The reason I feel confident about this change is that iOS already behaves this way. There are arguments for either style of dealing with errors, but I think the most important thing is to use the same style between platforms.

To be consistent, the options are:
1. Change iOS to generate hard errors when encoding fails -> this is a change that will break existing user code _at runtime_, probably crashing their apps
2. Change Android to fallback to empty strings like iOS does -> this will only break code that relied on exceptions being produced, which I would expect to be a very small minority, and it won't cause crashes

Lastly, about the `byte_error` and `wide_error`: this was my first approach, but it did not work. I think `wstring_convert` checks whether they're set by just checking if they're non-empty. Since we want to return empty strings, `wstring_convert` decides to throw instead. Cppreference is pretty unclear about this behavior, but here's a [godbold](https://godbolt.org/z/sW845devY) to demonstrate.

PR #176, opened 2024-07-08
Use official C++ 20 coroutine feature test macro

jb-gcx:
The djinni::Future coroutine support stops working at clang 17, see the reason here: https://stackoverflow.com/a/78027460/1480324

In this PR I use the updated the feature test macro as well as the outdated one.

### DIFF
--- a/support-lib/cpp/Future.hpp
+++ b/support-lib/cpp/Future.hpp
@@ -25,7 +25,7 @@
 #include <cassert>
 #include <exception>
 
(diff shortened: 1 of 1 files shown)


paulocoutinhox:
I think that you can remove comment "// __cpp_coroutines is outdated, __cpp_impl_coroutine is the standard feature test macro", no?

jb-gcx:
True, done 👍

LiFengSC (approved):

PR #177, opened 2024-07-10
Move result out of Future instead of copying it

jb-gcx:
I noticed that results are copied out of djinni::Futures instead of moved, probably due to a forgotten std::move.

Example: `auto ptr = djinni::Promise<std::unique_ptr<int>>::resolve(nullptr).get();`.
Before this it didn't work (called copy constructor of unique_ptr), now it does.

### Testing

I wanted to add a test to verify this works and doesn't break with future versions, but there are no dedicated C++ tests yet. I don't have experience with bazel, so I just added an XCTest to do this. I hope that's okay with you.

### DIFF
--- a/support-lib/cpp/Future.hpp
+++ b/support-lib/cpp/Future.hpp
@@ -56,13 +56,13 @@ template <typename T>
 struct ValueHolder {
     using type = T;
     std::optional<T> value;
-    T getValueUnsafe() const {return *value;}
+    T&& getValueUnsafe() && {return std::move(*value);}
 };
 template <>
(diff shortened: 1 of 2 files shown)


LiFengSC:
Thanks for the contribution.  I'm going to merge a slightly variant of your version, basically adding an overload for non-const callers:
```
    T getValueUnsafe() {return std::move(*value);}
```
the return type is a value type `T` rather than reference type `T&&` because I feel a bit uneasy to return references.

LiFengSC:
https://github.com/Snapchat/djinni/pull/180

PR #179, opened 2024-07-14
Fix extern type usage with non-nullable ptrs

trblunt:
When non-nullable pointers are set, externally defined interface types were not being treated the same as file-local interface types, causing missing import headers and incorrect pointer types to be used in C++ generated code.


LiFengSC (approved):
Thank you!

PR #187, opened 2024-08-12
Finish futures when their promise is broken

jb-gcx:
Accidentally discarding a `djinni::Promise` leads the associated future(s) to block forever in `get`. This is unrecoverable when it happens. Instead, I'm applying the behavior of `std::promise::~promise` to `djinni::Promise`: if the promise is not finished, we set a `BrokenPromiseException` into the shared state.

This is (kind of) a breaking change. But it only changes behavior for users, who rely on futures that never complete when the promise is discarded. In my opinion, that was buggy behavior.

### DIFF
--- a/support-lib/cpp/Future.hpp
+++ b/support-lib/cpp/Future.hpp
@@ -54,6 +54,12 @@ namespace djinni {
 template <typename T>
 class Future;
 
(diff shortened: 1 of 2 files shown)


LiFengSC (approved):
Thank you!

PR #195, opened 2025-01-17
Update .bazelversion to 7.2.1

nkoroste:


.bazelversion
```diff
@@ -1 +1 @@
-5.4.1
+7.2.1.29
```
nkoroste:
```suggestion
7.2.1
```

PattaFeuFeu:
The build fails because `--nouse_workers_with_dexbuilder` no longer exists with 7.2.1. If I get https://github.com/bazelbuild/bazel/issues/10241 right, it seems like that flag would no longer be necessary anyway?

Would it now be possible to get bazel to a newer version for Djinni?

rebwar101 (approved):

PR #196, opened 2025-04-30
Rename Composer:: -> Valdi::

clholgat:


paulocoutinhox:
How to use Valdi em mobile apps?

li-feng-sc:
Thanks,  we need to update the files under support-lib too.

clholgat:
I have updated the support-lib internally, is there something else that needs to be done in this repo?

li-feng-sc:
They should be the same.  Normally it's the other way round,  we make the change in the OSS repo, then pull the changes into the internal repo.   So here we can just copy the files from the internal repo back here to keep them in sync.

li-feng-sc:
> How to use Valdi em mobile apps?

The team is actively working on the open source release

li-feng-sc (approved):

PR #204, opened 2025-08-27
Added supports for constants in C codegen

rFlex:
This change adds supports for constants in C codegen. It generates getters that fetches the constants from C++, and performs the conversions appropriately. It supports primitives, strings, optionals, records etc...


li-feng-sc (approved):
lgtm

PR #217, opened 2026-01-14
Add a provider<> extension type

li-feng-sc:
|language | mapping|
| -------- | ------- |
|IDL | `provider<T>`|
| C++ | `djinni::Provider<T>` (alias of `std::function<T()>`)|
| Java | `com.google.common.base.Provider<T>`|
| ObjectiveC| `DJProvider<T>` (custom type, access target with `[provider get]`)|
|TypeScript | `Provider<T>` (alias of `() => T`)|
| Valdi | Same as TypeScript |
| Swift | `Provider<T>` (alias of `()->T`)| 

New tests are added for all languages, tests verify:
* Sending and receiving of primitive types wrapped in `provider`
* Sending and receiving of record types that contain `provider` members
* Sending and receiving of interface types wrapped in `provider`
* The factory method is not called before the target is requested
* The factory method is called when the target is requested
* The type works with a memoized provider (factory method only called once if memoized)


support-lib/cpp/Lazy.hpp
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+#pragma once
+
+#include <functional>
+
+namespace djinni {
+
+template<typename T>
+using Lazy = std::function<T()>;
```
lshire-sc: per our discussion on slack, let's just call these types "Supplier" since IMHO it's not a Lazy if it's not memoized.

support-lib/java/com/snapchat/djinni/Lazy.java
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.snapchat.djinni;
+
+import java.util.function.Supplier;
+
+/**
+ * Lazy<T> is just an alias for Supplier<T> in Java.
```
lshire-sc: Can we call this DjinniSupplier or something? IMHO it's not a lazy if it's not memoized

lshire-sc (approved):
the swift and objc implementations seem reasonable, though i'm much less familiar with the other languages. my main piece of feedback is that IMHO it's not a "lazy" type if it's not memoized. 

I'm fine with keeping this implementation as-is and referring to all of them as "suppliers" as you suggested, or alternately making a real Lazy type per language that does support memoization and thread safety (and would allow for checking if a type is initialized across languages.

either way this should be sufficient to pass dependencies to and from c++ without unwrapping them

li-feng-sc:
Renamed the type to 'provider<>'

